PR #10, opened 2015-08-14
Saving INF as text file to avoid CRLF issues.

bhouse-microsoft:
The INF file was saved as a UTF-16 file which GIT will see as binary for the most part.  However, we have specified that all INF files are text that should have CRLF conversions in the repository .gitattributes file.  When we clone the respository, the CRLF conversions are still being done to the UTF-16 INF file(s) causing them to become mangled.  Building the sample fails as the INF essentially appears to be a corrupted file.

This change converts the INF file back to a pure ANSI unmangled text file.

All other UTF-16 INF files in the upstream repository should also be fixed.


bhouse-microsoft:
There is another way to deal with this problem by introducing a local .gitattributes file.  I'll suggest that instead given that these files are being replicated from SD.

PR #11, opened 2015-08-14
Attributing UTF-16 INF files as text but with eol=LF to avoid corruption

bhouse-microsoft:
Some of our INF files are Unicode encoded.  Since we have a default rule that all INF files are text files that should use CRLF normalization on checkout, these files become corrupted.

To avoid corruption, we have created local exceptions for these specific files.  These Unicode INF files are attributed as text but with LF only normalization which will cause these files to remain untouched and not corrupted on checkout.


jmrossy:
Thanks Bart, but this is no longer needed.

PR #38, opened 2015-12-23
Avoid 0x80000003 on 32-bit Windows in msnmntr.

plujon:
The exception occurs in MonitorCoStreamFlowDeletion, but the bug is in MonitorCoCreateFlowContext.

In 32-bit Windows, the high bit of kernel mode pointers is set, and the pointer value is 0x80000000 or above.  Casting 0x80000000 to a UINT64 yields 0xffffffff80000000 because of sign extension.  ULongLongToULongPtr converts 0xffffffff80000000 to 0xffffffff.  Boom.

Cast using an intermediate uintptr_t to avoid the sign extension in the first place.

### DIFF
--- a/network/trans/msnmntr/sys/msnmntr.c
+++ b/network/trans/msnmntr/sys/msnmntr.c
@@ -384,7 +384,7 @@ cleanup:
       flowContext = NULL;
    }
 
-   return (UINT64) flowContext;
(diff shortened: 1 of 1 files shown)


m3liss4:
Thank you very much! I've been struggling for a day because whenever I started TraceView to capture Internet Explorer's events, the system crashed immediately.

PR #42, opened 2016-01-15
Update ecdisp.c to fix HIDP_SET_USAGE_VALUE_ARRAY

sjlevin:
The HIDP_SET_USAGE_VALUE_ARRAY case does not work because the ListLength is hardcoded to 0. Removing this line seems to make it work.


sjlevin:
I closed this pull request because I made a new one with another fix. See request #50

PR #50, opened 2016-02-26
Update hid\hclient

sjlevin:
hid\hclient has two issues that are fixed with these changes:

1. Extended Client Call HidP_SetUsageValueArray currently does nothing because its ListLength is hardcoded to 0

2. Continuous Asynchronous Reads no longer infinitely loops if a device returns a non-success code on an asynchronous read.


sjlevin:
Didn't work as expected. Closing.

PR #51, opened 2016-02-26
Update hid\hclient

sjlevin:
hid\hclient has two issues that are fixed with these changes:
1. Extended Client Call HidP_SetUsageValueArray currently does nothing because its ListLength is hardcoded to 0
2. Continuous Asynchronous Reads currently infinitely loops if a device returns a non-success code on an asynchronous read. This is especially bad for removable devices as the framework returns all asynchronous failure codes when they are removed.


wm1:
Typo. the commit 81bc1aeafa above is not related

wm1:
Fix is already in WDK version 1607.

Issue #49, opened 2016-02-18
Unable to save file to RAM disk drive. Drive 'R' not found.


tchaudev:
This issue is related to the [RAMDisk Storage Driver Sample](https://github.com/Microsoft/Windows-driver-samples/tree/master/storage/ramdisk). After building, and installing, the RAM disk driver I was not able to save a file on the RAM disk, because drive 'R' was not found. I followed the instructions provided in the README.md, and was able to build and install the RAM disk without an issue. Moreover, I could see the **WDF Sample RAM disk Driver** in the Device Manager, and the Ramdisk entries in the Registry. Is there something else I must do?

tchaudev:
Below is the full error message:
`Set-Location : Cannot find drive. A drive with the name 'R' does not exist.
At line:1 char:1
+ Set-Location R:
+ ~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (R:String) [Set-Location], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound,Microsoft.PowerShell.Commands.SetLocationCommand`

v-junyli:
We are closing out this issue that unfortunately have not been addressed for a long time.  Our apologies for not attending to the issue in a timely manner.  If you have data that issue remains and it is still important, then feel free to reopen.

Issue #44, opened 2016-01-20
RAMDisk Storage Driver Sample


marinkobabic:
Have installed the ramdrive on Windows 10 x64. I used the manual installation approach. The device was mounted and I have been able to see the drive R in explorer. When running the cmd console as administrator  and trying to switch to the drive R: "incorrect function" message is displayed.

tchaudev:
Did you ever figured out this issue? I'm having a similar problem, except I don't even see the drive 'R' in explorer. When using the Command Prompt I'm also getting "incorrect function", and a "DriveNotFoundException" when using PowerShell.

marinkobabic:
In the example FILE_SYSTEM_DISK is used. So when the drive gets the first request it tries to mount the volume. The mounting of the volume is not handled properly in the example and that is the reason we get this error message. 

On the web a have found the previous version of ramdisk for x86 operating system and this example is working as expected. It has been built for windows 2000/xp.

tchaudev:
Thanks for the info.

tchaudev:
I updated the RAMDisk driver sample so that drive "R:" can be accessed via command line. Please see pull request #53, if you're interested.

wzydhek:
I have the latest version and have it working correctly in my hyper-v machines, but I still get the incorrect function when using it on my host machine (also windows 10 x64).

cprogrammer1994:
I have a virtual machine with Windows x64. the driver installs properly with the devcon.exe. The `R:` drive appears. When I try to open it I get an "Access Denied" error. Running a cmd as administrator and running `R:` I get "Incorrect function."

What is the solution for this?

tchaudev:
See pull request #53

cprogrammer1994:
@tchaudev I forked your repo. I had some build errors in the inx file. I fixed them. It is not working either :(
I always test using a snapshot of a clean Windows install.

Did you test on Windows 10 x64?

tchaudev:
Yes, I tested on Windows 10 x64, but that was a long time ago. It is possible that the fix is out dated. That said, the problem was pretty simple. The reason why you get access denied is because the Ramdisk driver was not handling IOCTL queries that was introduced in Windows 10. If you update the code to properly handle those IOCTL, then everything should work.

cprogrammer1994:
Thank you! At least I know where to start. :)

chebotarevmichael:
@tchaudev, could you open source code of your sample for RAMdisk? It was not found in samples :(

[...]

PR #53, opened 2016-03-04
Added device IO control support for Windows 10

tchaudev:
Added support for the following device IO control codes:

- IOCTL_STORAGE_QUERY_PROPERTY
- IOCTL_DISK_GET_LENGTH_INFO
- IOCTL_MOUNTDEV_QUERY_DEVICE_NAME

Issues addressed: #49 and #44

### DIFF
--- a/storage/ramdisk/src/ramdisk.c
+++ b/storage/ramdisk/src/ramdisk.c
@@ -228,9 +228,6 @@ Return Value:
     size_t            bufSize;
     PDEVICE_EXTENSION devExt = QueueGetExtension(Queue)->DeviceExtension;
 
-    UNREFERENCED_PARAMETER(OutputBufferLength);
-    UNREFERENCED_PARAMETER(InputBufferLength);
-
     switch (IoControlCode) {
     case IOCTL_DISK_GET_PARTITION_INFO: {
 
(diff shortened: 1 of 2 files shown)


Smekalov:
tchaudev, is that helped and RAMDisk working as normal under Windows 10?

tchaudev:
**Is that helped and RAMDisk working as normal under Windows 10?**
This fix will get the ramdisk working under Windows 10 via PowerShell or CommandPrompt. I explain the issue more clearly here: http://stackoverflow.com/questions/35999668/windows-ramdisk-storage-driver-sample-drive-with-the-name-r-does-not-exist.

**Inside of manual for MS RAMDisk exists info about Package project but I not found it inside Solution. Is that normal?**
When I forked from Windows-driver-samples, the ramdisk solution never had the package project. Maybe it was added recently. That said, you don't need the package project to build, install, and run the WdfRamdisk project.

Smekalov:
Thanks. I see now. Explorer still not understand the driver on your PC too. I have compiled both versions one from MS and your fix - both not working well on my Windows 10 x64/VS2015/ WDK10. But I had one more problem - inf-file do not adds keys to Windows registry. Only when I added registry keys manually then x64-compiled driver was installed and started. Device Manager now have info about the driver and it looks as working normal in Device Manager but R: drive in File Explorer looks like Local Drive and have 0 size.

Smekalov:
I see now - the RAMDisk working only under administrator`s rights. If you have Total Commander installed - Run it  as Administrator. Then you`ll see working MS-RAMDRIVE as R: disk. Probably reason in x86/x64 compatibility of Windows 10. As I understand - necessary to use Wow64DisableWow64FsRedirection for x86-compiled driver - I have no other idea. :-)

Jichao:
thx for your work, i finally get ramdisk run on win 10.

tchaudev:
@Jichao thanks for the tip.

WEBAPIBUILDER:
I can't find the .inf file, is it missing?
It doesn't want to install correctly (on a Windows 2016 Server), any tips?

gkhanna79:
The missing INF file issue happens because ramdisk.inx is not included in the project. Add it to the _Driver Files_ section and rebuild. You will then run into two more errors about duplicate entries - just merge them into one and you will have the expected output files.

gkhanna79:
Interestingly, I was able to access the R: drive on a physical box but when I deploy the driver to Win10 VM (under HyperV), I get "Incorrect function" error as well. 

I will try out the PR and share an update.

gkhanna79:
I have confirmed that this PR resolved the _Incorrect Function_ issue. Thanks you @tchaudev.

gkhanna79:
Is there a reason this PR is not merged yet (blocked on owners to merge it)?

Smekalov:
Dear, is possible to compile this project under Visual Studio 2017 + WDK 1703? I even can not to open it.

Smekalov:
Dear, I tested RAMDisk sample under Windows 10 but found one trouble. Windows 10`s File Explorer defines RAMDisk drive as DVD RW Drive (R:). Any ideas how to change properties of RAMDisk drive from DVD RW to RAM DRIVE?

tchaudev:
@Smekalov, I couldn't get RAMDisk to work on Windows 10 file explorer when I posted this pull request, and I haven't touch it since. Try reading Jichao comment?

Smekalov:
Yes, the native sample have 'rights' problem. I have fixed it before Jichao message but I had other problems, that is a reason why I not sent this 'rights/security' fix here. Right now I have sources with your IO-fix, L"D:P(A;;GA;;;WD)" security and without FAT12 limitation. The driver working under MS Windows 10 x64. The driver possible to compile under Visual Studio 2015 + WDK 1703. Visual Studio 2017 still not supported WDK. I can sign the driver and install on my MS Windows 10 x64 system but I have some problems when I use it - for example File Explorer understand it like any sort of drive but not RAM drive.

tchaudev, I can send you changed sources but not know how to do this here.

P. S. I still interesting - why RAMDisk code, so usual for xNIX platform, still so secret for MS Windows? We had Win 3.11/95/XP/Me/2000/VISTA/7/8/8.1/10 but we never had normal system RAM disk with user interface like SUBST. So interesting thing still not available for usual users in time when PC have RAM from 6GB to ~.

tchaudev:
@Smekalov, maybe you could try opening an issue. I haven't touch this code in over a year, and don't have any system setup to build, or test, windows drivers.

HBelusca:
Asking in year 2024:
What's the current status of this PR (and the related issue https://github.com/microsoft/Windows-driver-samples/issues/44 )?

PR #103, opened 2017-01-02
added unicode support for devcon

FaraSeer:
Proposed changes because the output of the devcon for the russian version of windows was uninformative. For example, the output of "devcon.exe find usb\\*" command was as follows:

> USB\ROOT_HUB\4&1C7C1FAE&0                                   : ???????? USB-????????????
> USB\ROOT_HUB\4&3A448A3B&0                                   : ???????? USB-????????????
> USB\VID_09DA&PID_9090&MI_00\6&2FEDCB8B&0&0000               : USB-?????????? ?????
> 3 matching device(s) found.

### DIFF
--- a/setup/devcon/cmds.cpp
+++ b/setup/devcon/cmds.cpp
@@ -102,7 +102,7 @@ Return Value:
         for(dispIndex = 0;DispatchTable[dispIndex].cmd;dispIndex++) {
             if(DispatchTable[dispIndex].shortHelp) {
                 FormatToStream(stdout,DispatchTable[dispIndex].shortHelp,DispatchTable[dispIndex].cmd);
-                fputs("\n",stdout);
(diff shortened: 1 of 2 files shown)


wm1:
Test on Win10 v1607. Code page 437. After call `_setmode(_fileno(stdout), _O_WTEXT)`, the following console fonts display multiple (including CJK) languages correctly. Other fonts generally fail on CJK chars.

console-test

wm1:
However, there is app compatibility issue. I'm aware of some scripts that runs devcon, followed by findstr to filter the result. Note that findstr does not support Unicode, thus if devcon changes to output Unicode, findstr will find nothing.

Suggest to add an option for Unicode output, with default to off.

wm1:
Full commit is f36ae05fadd28e7f355297fd5f48177d9b68fbdc

PR #114, opened 2017-02-08
Resolve SAL warnings and suppress 4595

girishpattabiraman:
Resolve SAL annotations 
Ignoring 4595 and removing the project DB auto-generated file 
Ignore DB and opendb files. 

Warning 4595 can be resolved by locally implementing "operator new and delete", but the change has been reverted following guidelines from Audio team.


kfroe:
Resolve SAL warnings and suppress 4595

wm1:
commit 42cd81689c12, a982f7e243? Any insight on why suppress is chosen instead of new/delete override?

PR #118, opened 2017-03-10
Fix OACR code analysis warnings and update to include new metadata structure definitions and depend on new public ACPI definitions

saredd:


NaveenThum:
Does this compile?

AVStream/avscamera/mft0/sources.dep
```diff
@@ -1,21 +0,0 @@
-PUBLIC_PASS0_CONSUMES= \
```
NaveenThum: This doesn't need to be made into GitHub. Does it?

saredd: I am deleting that file as part of this change. Should I leave it in instead?

NaveenThum: My bad. Deletion is correct.

AVStream/avscamera/sys/Common.h
```diff
@@ -414,23 +414,9 @@ typedef enum _PIN_MODE
 
 } PIN_MODE, *PPIN_MODE;
```
NaveenThum: Ensure that new acpi.h is available in the released platform SDK with this definition (AFAIK, it is part of RS2 SDK, not RS1 SDK). Otherwise, we may need to retain AcpiPldRotation definition.

PR #129, opened 2017-04-13
devcon: Made DumpDeviceDriversCallback be __stdcall.

DavidEGrayson:
I've noticed the recent activity in this repository and was wondering if you would accept some pull requests for better MinGW support.  Here is the first one.  My commit message explains it pretty well.  Thanks!

### DIFF
--- a/setup/devcon/dump.cpp
+++ b/setup/devcon/dump.cpp
@@ -407,8 +407,7 @@ Return Value:
     return TRUE;
 }
 
-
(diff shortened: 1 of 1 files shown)


DavidEGrayson:
I also have several changes I'd like to make for usbview, but I'll wait to hear back from you before putting effort into making another pull request.

wm1:
Would it be better to use CALLBACK directly?

DavidEGrayson:
Yes, I think so, actually.  Thanks for the response; I'll update the pull request some time.

DavidEGrayson:
I changed the pull request and tested that I can still build devcon with MinGW.  I have not tested Visual Studio.

DavidEGrayson:
Thanks!

PR #147, opened 2017-06-13
ZeroMemory statistics data for retrieval

yentsanglee:
Reset memory so later memory comparison can succeed.


yentsanglee:
Reset memory so later memory comparison can succeed.

PR #163, opened 2017-08-15
Removed the second sections and merged

aktsuda:
When we built the original version of statbus.inx, we got the following errors:
error 1209: Section [destinationdirs] is defined multiple times.
error 1209: Section [sourcedisksfiles] is defined multiple times.
So, I removed these second sections and merged them to the first sections.


aktsuda:
Resolved in https://github.com/Microsoft/Windows-driver-samples/commit/bf93ae3e9d74d8863e23e10482da04d01c9fa947

PR #173, opened 2017-10-31
Added category filter metadata.

barrygolden:


banesttahp1:
I can't the system wants to or going to re-install Windows, but it also says, missing the media driver

PR #194, opened 2018-01-22
Fix segfault in FillDeviceInfo

jovermann:
The loop over all buttonCaps erroneously references valueCaps fields which seems to be a cut/copy/paste error. This causes a segfault for my setup.


AndrewYangMSFT (changes requested):
Please remove commit #2 (Make console output visible for hclient in command line mode) from Pull Request #194, as you have it in Pull Request #195 already.

PR #195, opened 2018-01-22
Make console output visible for hclient in command line mode

jovermann:
On Windows 7 using VS 2017 when using hclient no console output was visible. This patch makes the console output visible again.


AndrewYangMSFT:
@jovermann, do you have any insights on why it doesn't work on Win7 using VS2017? Is it the problem on Win7 or VS2017? Does the same binary (built using VS2017) work on Win8+?

PR #212, opened 2018-02-21
Added callback routine for SimBattEvtDeviceFileCreate

winbatt:
SimBattEvtDeviceFileCreate callback routine has been added, when sample(miniclass) registered with class driver then windows kernel should be able to open a handle with this new battery miniclass instance. In absence of this function Class/miniclass work properly but from windows kernel prospective file handle on this driver fails.


wm1:
suppressed by new #213

PR #214, opened 2018-02-27
Added .yaml files for samples portal 

barrygolden:
Added YAML manifest files to each sample. These are required by the samples portal.


wpetrol:
it sucks all the effort

PR #217, opened 2018-03-01
Serial driver bug fix.

FrancoisGuimond:
If a request has a cancel routine, the cancel routine need to be cleared before completion otherwise the queue get corrupted.

### DIFF
--- a/serial/serial/utils.c
+++ b/serial/serial/utils.c
@@ -213,10 +213,19 @@ Return Value:
 
             reqContext = SerialGetRequestContext(oldRequest);
 
-            SerialCompleteRequest(oldRequest,
-                                  reqContext->Status,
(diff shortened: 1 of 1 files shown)


FrancoisGuimond:
Hummm,

Maybe my explanation was not clear, but that was a fix for a BSOD in a interrupt storm caused by the serial cable being wrongly electrically connected.  Now it has been over a year since I helped my colleague debug that BSOD, but I remember in the memory dump, one of the WDFQueue owning a WDFREQUEST that just was completed...  After a couple of debug and trial days, we figured that adding that call to SerialClearCancelRoutine() fixed it. 

I agree that it happened in an unexpected setting, but I still think there is a bug in that driver.

PR #246, opened 2018-05-16
Remove sdiomars & ramdisk samples

NeoAdonis:


mjsabby:
Why was this removed? https://docs.microsoft.com/en-us/windows-hardware/drivers/wdf/sample-kmdf-drivers still links to it.

barrygolden:
Thanks for catching that residual link to the removed RAMDisk sample in the **Sample KMDF Drivers** topic. 
I removed the link on Monday 6/4/2018.

PR #278, opened 2018-08-27
filesys/cdfs: properly check for legal names in CdIsLegalName()

HeisSpiter:
Up to now, it was working by chance. Indeed, due to the invalid
ASCII check performed before calling FsRtlIsAnsiCharacterLegalHpfs(), the
macro is improperly called and overruns the FsRtlLegalAnsiCharacterArray
buffer. Fortunately, up to now, right after that buffer in kernel binary
there are strings which are more or less consistent with the flags that
are expected by the macro, causing a decent behavior of
FsRtlIsAnsiCharacterLegalHpfs() even for extended ASCII characters
(whereas FsRtlIsAnsiCharacterLegalHpfs() is only designed for ASCII
characters). But this is a totally out of control and wrong behavior.
A single change in the way the kernel was built could have caused the
CDFS driver not to work as previously.

I have made the choice to allow any extended ASCII character as done
for the unicode characters. This is a good compromise to avoid drastic
regressions for users having extended ASCII characters in their CD
file names.

### DIFF
--- a/filesys/cdfs/namesup.c
+++ b/filesys/cdfs/namesup.c
@@ -404,7 +404,13 @@ Return Value:
          Wchar < Add2Ptr( FileName->Buffer, FileName->Length, PWCHAR );
          Wchar++) {
 
-        if ((*Wchar < 0xff) &&
+        //
+        // Check whether ASCII characters are legal.
+        // We will consider the rest of the characters
+        // (extended ASCII and unicode) as legal.
+        //
+
+        if ((*Wchar < 0x80) &&
             !FsRtlIsAnsiCharacterLegalHpfs( *Wchar, FALSE ) &&
             (*Wchar != L'"') &&
             (*Wchar != L'<') &&


darwin-msft:
Thanks for looking at the code! However I think the CDFS code is correct and this PR is not necessary. 

The first thing CDFS does is verify that Wchar is representable in a char (no lost high order bits) by checking if it is less than 0xff.

>         if ((*Wchar < 0xff) &&
>             !FsRtlIsAnsiCharacterLegalHpfs( *Wchar, FALSE ) &&

Then it calls the macro FsRtlIsAnsiCharacterLegalHpfs (in ntifs.h) which is:

> #define FsRtlIsAnsiCharacterLegalHpfs(C,WILD_OK) (                 \
>     FsRtlTestAnsiCharacter((C), TRUE, (WILD_OK), FSRTL_HPFS_LEGAL) \
> )

If you look at the code for FsRtlTestAnsiCharacter (in ntifs.h) which is:

> #define FsRtlTestAnsiCharacter(C, DEFAULT_RET, WILD_OK, FLAGS) (            \
>         ((SCHAR)(C) < 0) ? DEFAULT_RET :                                    \
>                            FlagOn( LEGAL_ANSI_CHARACTER_ARRAY[(C)],         \
>                                    (FLAGS) |                                \
>                                    ((WILD_OK) ? FSRTL_WILD_CHARACTER : 0) ) \
> )

You'll note the first thing it does is cast the char C to a signed char (SCHAR). If ((SCHAR)(C) < 0), it returns DEFAULT_RET (in this case TRUE), instead of overflowing the array.

Thus I think there is no overflow here, and hence this PR should not be accepted into Microsoft:master.

filesys/cdfs/namesup.c
```diff
@@ -404,7 +404,13 @@ Return Value:
...
          Wchar++) {
 
-        if ((*Wchar < 0xff) &&
+        //
+        // Check whether ASCII characters are legal.
+        // We will consider the rest of the characters
+        // (extended ASCII and unicode) as legal.
+        //
+
+        if ((*Wchar < 0x80) &&
```
darwin-msft: Thanks for looking at the code! However I think the CDFS code is correct and this PR is not necessary. 

The first thing CDFS does is verify that Wchar is representable in a char (no lost high order bits) by checking if it is less than 0xff.

>         if ((*Wchar < 0xff) &&
>             !FsRtlIsAnsiCharacterLegalHpfs( *Wchar, FALSE ) &&

Then it calls the macro FsRtlIsAnsiCharacterLegalHpfs (in ntifs.h) which is:

>     #define FsRtlIsAnsiCharacterLegalHpfs(C,WILD_OK) (                 \
>         FsRtlTestAnsiCharacter((C), TRUE, (WILD_OK), FSRTL_HPFS_LEGAL) \
>     )

If you look at the code for FsRtlTestAnsiCharacter (in ntifs.h) which is:

>     #define FsRtlTestAnsiCharacter(C, DEFAULT_RET, WILD_OK, FLAGS) (            \
>             ((SCHAR)(C) < 0) ? DEFAULT_RET :                                    \
>                                FlagOn( LEGAL_ANSI_CHARACTER_ARRAY[(C)],         \
>                                        (FLAGS) |                                \
>                                        ((WILD_OK) ? FSRTL_WILD_CHARACTER : 0) ) \
>     )

You'll note the first thing it does is cast the char C to a signed char (SCHAR). If ((SCHAR)(C) < 0), it returns DEFAULT_RET (in this case TRUE), instead of overflowing the array.

Thus I think there is no overflow here, and hence this PR should not be accepted into Microsoft:master.

HeisSpiter:
My mistake! You're totally right.
Thanks for your detailed review, and sorry for the noise!

PR #303, opened 2018-10-19
Fix changing and reading FAT labels with leading 0xE5 byte in fastfat.sys

pali:
FAT Volume Label is stored in the root directory entry and therefore when
setting new Volume Label with leading 0xE5 byte it first needs to be
converted to 0x05 byte. This step is also required by Microsoft FAT
specification (fatgen103.doc) and also applied for all other files and
directories. But not for volume label yet. First byte 0xE5 in directory
entry means that entry is removed.

Similarly, when reading Volume Label from the root directory, leading byte
0x05 first needs to be converted to 0xE5 and then later processed. This is
reverse operation as described in previous paragraph.

FAT Volume Label is converted from Unicode to bytes according to OEM code
page. OEM code page is configured in Language for non-Unicode applications
settings in Country/Region settings.

When it is set to English (United Kingdom), then OEM code page is 850. And
in 850 is Unicode LATIN CAPITAL LETTER O WITH TILDE represented as 0xE5.

Therefore this pull request fixes setting FAT label "Õ" when regional
settings are set to English (United Kingdom).

### DIFF
--- a/filesys/fastfat/fsctrl.c
+++ b/filesys/fastfat/fsctrl.c
@@ -1411,22 +1411,34 @@ Return Value:
 
         if (Dirent != NULL) {
 
+            UCHAR OemBuffer[11];
             OEM_STRING OemString;
             UNICODE_STRING UnicodeString;
+
+            OemString.Buffer = (PCHAR)&OemBuffer[0];
+            OemString.MaximumLength = 11;
+
+            RtlCopyMemory( OemString.Buffer, Dirent->FileName, 11 );
+
+            //
+            //  Translate the first character from 0x5 to 0xe5.
+            //
+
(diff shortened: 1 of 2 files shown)


pali:
@NeoAdonis or somebody else, can you review this pull request?

pali:
@NeoAdonis ping

pali:
@NeoAdonis just gentle reminder of this PR, could you please look at it?

DragonbornSR:
> @NeoAdonis just gentle reminder of this PR, could you please look at it?

Let me find someone

pali:
@DragonbornSR Thank you! Let me know if you need some more details of explanation of these changes.

pali:
@NeoAdonis Thanks!

Just a one question: As fastfat.sys is also included in Windows itself, is this change going to be synchronized also into Windows code?

PR #322, opened 2018-12-12
Fix X86 support; Add ARM support.

JoelCorley:
Remove broken X86-specific DMA code.
Add project support for ARM & ARM64 targets.


avstream/avshws/capture.cpp
```diff
@@ -1079,7 +1079,7 @@ Routine Description:
            selected in order to change the format for the pin.  In this case,
            OldFormat will not be NULL.
 
-    Validate that the format is acceptable and perform the actions necessary
+    Validate that the format is acceptible and perform the actions necessary
```
NeoAdonis: acceptable

PR #363, opened 2019-04-13
Update README.md of PrinterExtensionSample

hsebs:
The link of "Building a Driver" is not works.
Update all URLs to "docs.microsoft.com".


hsebs:
Fixed in #424

PR #380, opened 2019-06-05
FastFAT: Check DBG with #if, not #ifdef

SergeGautherie:


HBelusca:
@microsoft/filesystems Could you please review this PR?

PR #393, opened 2019-08-01
Fix Avstream sample build for older Windows versions

Sudhaushu:


avstream/avscamera/mft0/stdafx.h
```diff
@@ -16,9 +16,11 @@
 #endif
 
 #include "targetver.h"
+#if (NTDDI_VERSION <= NTDDI_WIN7)
```
JoelCorley: I assume you're leaving out the deprecation message because we've not yet release Vibranium?

Sudhaushu: yes

---
In reply to: [309897773](https://github.com/microsoft/Windows-driver-samples/pull/393#discussion_r309897773) [](ancestors = 309897773)

PR #431, opened 2019-10-25
NULL pointer exception errors fixed

JinZhuXing:
I think, at this point, pIUsbTargetDevice is NULL.
So it will be used as m_pIUsbTargetDevice variable.

### DIFF
--- a/biometrics/driver/Device.cpp
+++ b/biometrics/driver/Device.cpp
@@ -569,14 +569,14 @@ Return Value:
 
     if (SUCCEEDED(hr)) 
     {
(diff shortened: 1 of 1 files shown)


biometrics/driver/Device.cpp
```diff
@@ -637,11 +637,12 @@ Return Value:
                     else if ( UsbdPipeTypeBulk == pIUsbPipe->GetType() )
                     {
                         m_pIUsbInputPipe = pIUsbPipe;
+						m_pIUsbInterruptPipe = pIUsbPipe;
```
SergeGautherie: Use spaces, not tabs.

PR #442, opened 2019-11-23
SFloppy: RtlStringCbPrintf?() need bytes, not chars

SergeGautherie:


SergeGautherie:
Updated commit comment.

PR #458, opened 2020-02-16
Fix build failure for file encoding.

yangrongwei:
Replacing Unicode characters with ASCII charaters.

audio\sysvad\EndpointsCommon\MiniportAudioEngineNode.cpp(1,1): error C2220: the following warning is treated as an error
audio\sysvad\EndpointsCommon\MiniportAudioEngineNode.cpp(1,1): warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss


jasonknichel:
I believe this has been taken care of with https://github.com/microsoft/Windows-driver-samples/pull/784.  Closing this PR

PR #485, opened 2020-04-25
classpnp: Use ExAllocatePoolZero instead of ExAllocatePoolWithTag

amaczugowski:


storage/class/classpnp/src/classp.h
```diff
@@ -26,6 +26,7 @@ Revision History:
 --*/
 
 #define RTL_USE_AVL_TABLES 0
+#define POOL_ZERO_DOWN_LEVEL_SUPPORT
```
mannanal, line 29: Drew, May I know why this is defined?

amaczugowski: Good question. It's a requirement to use the new allocation functions on versions earlier than 20H1.

Source:
https://osgwiki.com/wiki/Using_Pool_Zeroing_APIs#What_if_my_code_needs_to_run_on_downlevel_platforms.3F

msscolee (approved):

PR #493, opened 2020-05-01
Added configuration setting steps in Updated README.md 

Chao0000:
I added statements about how to change Platform Toolset of the projects in Step 2-i.
When I used VS2017 and VS2019 to try to build the projects, Driver Files wasn't shown in the menu. 
So I think it is necessary to add these steps for people who download these good sample projects.
Could you help me check what I added is right or not?
Thank you

Best Regard,
Chao


barrygolden:
Thanks for the pull request @Chao0000!

Chao0000:
Hello @barrygolden , thanks for your review. It my pleasure!

PR #504, opened 2020-05-26
Removed nuget package link, copied wil header files and link directly to solve lab machine test failed issue.

strongstrongyiwen:
Removed nuget package link, copied wil header files and link directly to solve lab machine test failed issue.
@JoelCorley 
@saredd


strongstrongyiwen:
deleted this "copy header files into project" solution. provided the packages folder which contains nuget package to the test team, could build clean without the need to restore package.

PR #513, opened 2020-06-08
Add multimon support to sample driver

zeusk:


zeusk:
Also fixes build for ARM64

PR #524, opened 2020-06-19
toastDrv/kmdf/filter/sideband: Ensure that queue event callbacks are run at PASSIVE_LEVEL

ColinFinck:
According to https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdfio/nc-wdfio-evt_wdf_io_queue_io_device_control, an `EVT_WDF_IO_QUEUE_IO_DEVICE_CONTROL` callback function may be called at any IRQL <= DISPATCH_LEVEL. It states especially:

> The EvtIoDeviceControl callback function can be called at IRQL <= DISPATCH_LEVEL, unless the ExecutionLevel member of the device or driver's WDF_OBJECT_ATTRIBUTES structure is set to WdfExecutionLevelPassive.

Not doing that is among the three most common mistakes when developing KMDF drivers according to https://www.osr.com/nt-insider/2015-issue3/common-kmdf-errors/

However, the toastDrv/kmdf/filter/sideband example currently doesn't specify any ExecutionLevel.
Instead, it just silences the related warning.

This PR adds the initialization of a WDF_OBJECT_ATTRIBUTES structure with ExecutionLevel set to WdfExecutionLevelPassive, which is then passed to WdfIoQueueCreate.

If this change is fine, the same fix should be applied to other sample drivers as well.
If WdfExecutionLevelPassive is deterministically not needed or even makes things worse, I would like to hear an explanation and have all related documentation adjusted. Currenlty, it feels like not using WdfExecutionLevelPassive is a serious mistake.

### DIFF
--- a/general/toaster/toastDrv/kmdf/filter/sideband/filter.c
+++ b/general/toaster/toastDrv/kmdf/filter/sideband/filter.c
@@ -375,6 +375,7 @@ Return Value:
     BOOLEAN                     bCreate = FALSE;
     NTSTATUS                    status;
     WDFQUEUE                    queue;
+    WDF_OBJECT_ATTRIBUTES       queueAttributes;
     DECLARE_CONST_UNICODE_STRING(ntDeviceName, NTDEVICE_NAME_STRING) ;
     DECLARE_CONST_UNICODE_STRING(symbolicLinkName, SYMBOLIC_NAME_STRING) ;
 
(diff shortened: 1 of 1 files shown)


ColinFinck:
Bump. Is anybody going to look at this?

ColinFinck:
Bump again

PR #534, opened 2020-08-22
fix bug in PackReport

ebadger:
Data is pointer to array of HID_DATA structure.
Function advances pointer as it packs the report buffer.
At end of process in success condition, the function attempts to walk the array a second time and update the IsDataSet member for the HID_DATA elements that were packed.

What it's actually doing in the second pass is writing unexpected memory locations.
Fix is to store the head of the array and move back before walking again to set the bits.


barrygolden:
@ebadger, PR is merged. Thanks!

PR #581, opened 2021-02-25
[vhidmini] Delete unreferenced DEVICE_CONTEXT struct DefaultQueue member

forderud:
This queue doesn't seem to be used for anything. Therefore, proposing to remove it to simplify the sample project.


forderud:
Closing PR, since the proposal appears to break the driver.

PR #587, opened 2021-03-13
input/layout: Fix settings application crash when loading selecting layout

### LINKED ISSUE #433: kbdus crashes settings/shell
I’m having an issue similar to #245. I successfully compiled kbdus.dll from input/layout/kbdus using `msbuild /p:Configuration=Release /p:Platform=x64` and the Enterprise Windows Driver Kit (WDK) build environment (ISO) Version 19h1_release_svc_prod3.18362.84 on Windows 10 Pro 1903 build 18362.449. Then I copied kbdus.dll into system32\kbdus2.dll, registered it:

```reg
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Keyboard Layouts\c0000407]
"Layout Text"="English (custom)"
"Layout File"="kbdus2.dll"
"Layout Display Name"="Some display name"
```

and tried to activate it using the control panel, which crashes after selecting the entry from the list. After rebooting the system the shell keeps crashing when logging in. Deleting kbdus2.dll fixes this.

It’s likely the build scripts are incorrect and lacking some flags. Comparing the built kbdus.dll to the files present in system32 it looks like it pulls in other dll’s like

- KERNEL32.dll
- api-ms-win-crt-runtime-l1-1-0.dll
- api-ms-win-crt-heap-l1-1-0.dll
- api-ms-win-crt-string-l1-1-0.dll
- api-ms-win-crt-stdio-l1-1-0.dll
- api-ms-win-crt-convert-l1-1-0.dll

when it shouldn’t.

davidebeatrici:
The first `AdditionalOptions` entry was not taken into account by Visual Studio, but the linker did.

Fixes #433.


davidebeatrici:
Turns out that:

1. My changes actually do fix the crash, but the layout is not actually loaded as it doesn't appear in the language bar and is not active.
2. #588 is ironically the real fix.

I have no clue why it appeared to be working at first. I believe I accidentally used a layout ID that was already assigned to another layout and as a result I tested the wrong one.

Sorry.

PR #598, opened 2021-03-28
uvcview.c: no console in quiet mode

imacovei:
If the /q command line switch is set, then don't initialize the console.
This is useful when the program is called from external components
and want to avoid annoying flashing of the console

Use a global boolean to store the quite mode


imacovei:
Found a bug: the  xml files aren't generated any more

PR #628, opened 2021-06-02
Update sha256 digest algo for network/ndis/extensions solution

rasunder-ms:
Updated sxbase.vcxproj
msforwardnext.vcxproj
mspassthroughext.vcxproj


wm1:
Already fixed as part of #656

PR #636, opened 2021-06-16
Changed ExAllocatePool to ExAllocatePool2 with tag

heqianw:
### Why this change is needed?

Due to the `ExAllocatePool` function being deprecated, we needed to update the API call to the ExAllocatePool2 function. The reason behind this change is that `ExAllocatePool` allocates memory that is uninitialized, users need to zero the memory before making it visible to users (Avoid leaking kernel-mode contents). In comparison, the memory that `ExAllocatePool2` allocates is already zero initialized; the security risk of disclosing information is not present with this function call.

### How was this change implemented?

The change was implemented by changing `ExAllocatePool` to a `ExAllocatePool2` function call. This was done with the help of the following APIs: [`ExAllocatePool`](`https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool`) and [`ExAllocatePool2`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2)

### How was this tested?

The driver was built using Visual Studio & WDK.

### DIFF
--- a/bluetooth/serialhcibus/Io.h
+++ b/bluetooth/serialhcibus/Io.h
@@ -50,6 +50,8 @@ Revision History:
 
 #define BUFFER_AND_SIZE_ADJUSTED(Buffer, Size, SegmentCount, Increment) {Buffer += Increment; Size -= Increment; SegmentCount += Increment;}
 
+#define POOLTAG_BTHSERIALHCIBUSSAMPLE 'htbw'
+
 #include <PSHPACK1.H>
 
 //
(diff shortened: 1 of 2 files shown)


heqianw:
@tobafo19 @wm1 Hello, would it be possible to approve this pull request? The changes were also completed internally. Seeing as you approved the changed to the other PRs, would it be possible to approve this one as well? Thanks

PR #668, opened 2021-11-09
Initialize local variables

matwilli:
Local variables may not be initialized prior to first access; initialize them at definition


AndrewYangMSFT (approved):
Having variables initialized is good. However, failure of getting windows coordinates may indicate invalid window handles, which may need separate investigation.

PR #696, opened 2022-03-03
[network\trans] ExAllocatePoolWithTag changing to ExAllocatePoolZero

houha2:
Per new security requirements the ExAllocatePoolWithTag are no longer secure using ExAllocatePoolZero instead of ExAllocatePool2 because these samples are downlevel drivers. Compiled with Visual Studio 2019. 

Also adds PnpLockdown = 1 to driver Inf files to resolve a compiler warning

### DIFF
--- a/network/trans/ddproxy/sys/DD_proxy.c
+++ b/network/trans/ddproxy/sys/DD_proxy.c
@@ -114,8 +114,6 @@ DDProxyFlowEstablishedClassify(
       goto Exit;
    }
 
-   RtlZeroMemory(flowContextLocal, sizeof(DD_PROXY_FLOW_CONTEXT));
-
    flowContextLocal->refCount = 1;
    flowContextLocal->flowType = (DD_PROXY_FLOW_TYPE)(filter->context);
    flowContextLocal->addressFamily = 
@@ -346,8 +344,6 @@ DDProxyClassify(
       classifyOut->rights &= ~FWPS_RIGHT_ACTION_WRITE;
       goto Exit;
    }
-
-   RtlZeroMemory(packet, sizeof(DD_PROXY_PENDED_PACKET));
 
    NT_ASSERT(flowContextLocal != NULL);
 
--- a/network/trans/ddproxy/sys/DD_proxy.c
+++ b/network/trans/ddproxy/sys/DD_proxy.c
@@ -102,7 +102,7 @@ DDProxyFlowEstablishedClassify(
 #endif /// (NTDDI_VERSION >= NTDDI_WIN7)
    UNREFERENCED_PARAMETER(flowContext);
 
-   flowContextLocal = ExAllocatePoolWithTag(
+   flowContextLocal = ExAllocatePoolZero(
                         NonPagedPool,
                         sizeof(DD_PROXY_FLOW_CONTEXT),
                         DD_PROXY_FLOW_CONTEXT_POOL_TAG
@@ -334,7 +334,7 @@ DDProxyClassify(
       goto Exit;
    }
 
-   packet = ExAllocatePoolWithTag(
+   packet = ExAllocatePoolZero(
                      NonPagedPool,
                      sizeof(DD_PROXY_PENDED_PACKET),
                      DD_PROXY_PENDED_PACKET_POOL_TAG
@@ -413,7 +413,7 @@ DDProxyClassify(
       {
          NT_ASSERT(inMetaValues->controlDataLength > 0);
 
-         packet->controlData = ExAllocatePoolWithTag(
+         packet->controlData = ExAllocatePoolZero(
                                  NonPagedPool,
                                  inMetaValues->controlDataLength,
                                  DD_PROXY_CONTROL_DATA_POOL_TAG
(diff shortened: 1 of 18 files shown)


network/trans/ddproxy/sys/DD_proxy.c
```diff
@@ -102,7 +102,7 @@ DDProxyFlowEstablishedClassify(
 #endif /// (NTDDI_VERSION >= NTDDI_WIN7)
    UNREFERENCED_PARAMETER(flowContext);
 
-   flowContextLocal = ExAllocatePoolWithTag(
+   flowContextLocal = ExAllocatePoolZero(
```
wm1, line 105: Remove RtlZeroMemory in line 117 since you're now calling ExAllocatePoolZero. Do the same for all other changes.

Also, have you done the following: "To run on versions of Windows prior to Windows 10 version 2004, the driver must define POOL_ZERO_DOWN_LEVEL_SUPPORT and call ExInitializeDriverRuntime before calling this function. "

houha2: Will do - I'll double the check the samples call ExInitializeDriverRuntime as well, but the dd_proxy sample does call it in DriverEntry() in DD_drv.c

Would the POOL_ZERO_DOWN_LEVEL_SUPPORT be defined in the vcxproj file or is it defined in source?

wm1: re: POOL_ZERO_DOWN_LEVEL_SUPPORT
If put into vcxproj, each $(Configuration)|$(Platform) needs to be updated, which is a little bit of complex. IMO it is simpler to define it before including <ntddk.h> or <wdm.h>.

houha2: Thanks just pushed an update - adding those lines

network/trans/ddproxy/sys/ddproxy.vcxproj
```diff
@@ -46,7 +46,7 @@
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <TargetVersion>Windows10</TargetVersion>
     <UseDebugLibraries>False</UseDebugLibraries>
-    <DriverTargetPlatform>Universal</DriverTargetPlatform>
+    <DriverTargetPlatform>Desktop</DriverTargetPlatform>
```
wm1, line 49: Make sure all flavors build successfully. For example, other places of DriverTargetPlatform need be updated to Desktop too.

network/trans/stmedit/sys/stmedit.vcxproj
```diff
@@ -30,7 +30,7 @@
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <TargetVersion>Windows10</TargetVersion>
     <UseDebugLibraries>False</UseDebugLibraries>
-    <DriverTargetPlatform>Universal</DriverTargetPlatform>
+    <DriverTargetPlatform>Desktop</DriverTargetPlatform>
```
wm1, line 33: make sure all flavors build successfully. for example, x64/debug currently fail because TargetVersion is not defined.

network/trans/inspect/sys/inspect.c
```diff
@@ -22,7 +22,7 @@ Copyright (c) Microsoft Corporation. All rights reserved
 
 --*/
 
-
+#define POOL_ZERO_DOWN_LEVEL_SUPPORT
```
wm1, line 25: make sure all flavors build successfully. for example, inspect.vcxproj line 26, <Platform> should be set to a valid value, e.g. x64.

houha2: Just tested all flavors build for all samples being modified

network/trans/ddproxy/sys/ddproxy.vcxproj
houha2: See below comment - all flavors build for all samples edited

network/trans/stmedit/sys/stmedit.vcxproj
houha2: Thanks for pointing this out - it has been addressed see other comments

network/trans/stmedit/sys/stmedit.inf
```diff
@@ -23,17 +24,20 @@
     DefaultDestDir      = 12                              ; %WinDir%\System32\Drivers
     StmEdit.DriverFiles = 12                              ; %WinDir%\System32\Drivers
 
-[DefaultInstall]
+[DefaultInstall.NTamd64]
```
wm1, line 27: why hardcode amd64 here? doing so will cause driver fail to install on x86.

houha2: When building I got a warning: "warning 1420: [DefaultInstall]-based INF cannot be processed as Primitive.", and then I used this https://docs.microsoft.com/en-us/windows-hardware/drivers/develop/creating-a-primitive-driver to do the architecture decoration. Will add the x86 ones.

wm1: the recommended approach is to use "NT$ARCH$", which will be expanded by stampinf during build.











The correct a

PR #698, opened 2022-03-10
[serenum, wmisamp] ExAllocatePoolWithTag changing to ExAllocatePoolZero

thpthp1:
Per new security requirements the ExAllocatePoolWithTag are no longer secure using ExAllocatePoolZero instead of ExAllocatePool2 because these samples are down level drivers. Builds with Windows 11 EWDK with Visual Studio Build Tools 16.9.2.

Also adds PnpLockdown = 1 to driver Inf files to resolve a compiler warning


serial/serenum/string.c
```diff
@@ -384,7 +384,7 @@ Return Value:
    multiLength += sizeof(WCHAR);
 
    MultiString->MaximumLength = (USHORT)multiLength;
-   MultiString->Buffer = ExAllocatePoolWithTag(PagedPool, multiLength,SERENUM_POOL_TAG);
+   MultiString->Buffer = ExAllocatePool2(POOL_FLAG_PAGED, multiLength,SERENUM_POOL_TAG);
```
wm1, line 387: Use the Zero API like the other places

PR #705, opened 2022-03-19
Replace Deprecated API

pralater:
Replaced ExAllocatePoolWithTag with ExAllocatePool2 to fix a memory initialization security vulnerability.

Removed RtlZeroMemory because memory is zero initialized by ExAllocatePool2.


pralater:
made additional fix to code using outdated definitions

PR #716, opened 2022-04-20
Check success of StringFromCLSID before printing string (#1)

microkatz:
* Check success of StringFromCLSID before printing string
* initialized wchar ptr to null


NaveenThum (approved):
:shipit:

PR #745, opened 2022-06-17
NameChanger - Update INF Errors during Build.

rakendrathapa:
Building the NameChanger project showed following errors:

> "[DefaultInstall]-based INF cannot be processed as Primitive"

Made changes as mentioned in the link: https://docs.microsoft.com/en-us/windows-hardware/drivers/develop/creating-a-primitive-driver

There are other drivers which show the same issue. If this is merged, will make changes for others as well.


rakendrathapa:
Adding code changes for adding functions to the Text Allocation using Pragmas.

PR #767, opened 2022-08-10
Use includes/needs for wudfbiousbsample

MaliParag:


emilyhobby:
For multiple Include/Needs, use a single directive, i.e.

Include=WINUSB.INF, WUDFRD.INF
Needs=WINUSB_LowerFilter.NT,WUDFRD.NT

emilyhobby (approved):

PR #778, opened 2022-08-31
Include/Needs inbox services for hidusbfx2, vhidmini, umdf_filter_umdf

emilyhobby:
Updated 8 sample drivers to Include/Needs inbox services as functional and filter drivers instead of copying the AddService directives from the inbox drivers.

The minimum OS version to Include/Needs winusb service as an upper or lower filter is Windows 11 22H2.
The minimum OS version to Include/Needs mshidkmdf service is Windows 11 21H2.

Built each sample and confirmed InfVerif error 2087 was present. Updated the samples and rebuilt, confirmed the error cleared.


Shyamal-V (commented):
@wm1 can you please review?

wm1 (approved):

PR #802, opened 2022-11-09
Do not use deprecated API ExAllocatePool

taksaku:
Using ExAllocatePool2 instead of deprecated API ExAllocatePool

### DIFF
--- a/smartcrd/pscr/pscrnt.c
+++ b/smartcrd/pscr/pscrnt.c
@@ -406,8 +406,8 @@ PscrRegisterWithSmcLib(
     SmartcardExtension = &DeviceExtension->SmartcardExtension;
 
    //   allocate the reader extension
-    ReaderExtension = ExAllocatePoolWithTag(
-                                    NonPagedPoolNx,
+    ReaderExtension = ExAllocatePool2(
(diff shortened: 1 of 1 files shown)


smartcrd/pscr/pscrnt.c
```diff
@@ -406,8 +406,8 @@ PscrRegisterWithSmcLib(
     SmartcardExtension = &DeviceExtension->SmartcardExtension;
 
    //   allocate the reader extension
-    ReaderExtension = ExAllocatePoolWithTag(
-                                    NonPagedPoolNx,
+    ReaderExtension = ExAllocatePool2(
```
jasonknichel, line 409: By default, unless you pass POOL_FLAG_UNINITIALIZED, ExAllocatePool2 will zero the memory it returns, so you should not need this anymore:
    RtlZeroMemory( ReaderExtension, sizeof( READER_EXTENSION ));

jasonknichel:
We request that the majority of changes happen into the develop branch and not main.  I actually already made this change in the develop branch in early October: https://github.com/microsoft/Windows-driver-samples/pull/788 .  Closing this PR.  When we merge develop into main next, main will pick up this fix.

PR #803, opened 2022-11-16
Fix driver isolation violations 1322 and 1323 for netvmini (60, 620, 630) drivers.

kahashimoto-ms:
Fixed driver isolation violations 1322 and 1323 for netvmini60, netvmini620, and netvmini630. 
Verified by installing on a VM and pinging the devices.


jasonknichel:
@kahashimoto-ms - Thanks for this PR!  We took a look and the changing of the destination directory of the sys files look fine but we have a request for the TextModeFlags registry value.  This looks like something that text mode setup (Win2k and earlier way of installing Windows NT based OSes) would have looked at.  We don't believe anything in the OS has read it for a long time, and if it was still reading it, it would be looking for it in the old location and not the new location you moved it to.  Instead of moving the location of the TextModeFlags registry value, can you please remove it from the INF altogether? Thanks!

jasonknichel (changes requested):
Please see my other comment for things to address in this PR.

jasonknichel (approved):

PR #804, opened 2022-11-16
Netlwf iso violations

kahashimoto-ms:


jasonknichel:
@kahashimoto-ms - running 'infverif /v /w /stampinf' on your modified INF still shows a "Windows Driver" violation.  Have you taken a look at this violation?  Is it possible to address it currently or does the device stack not support moving the file?

ERROR(1322) in S:\git\testDrv\network\ndis\filter\netlwf.inf, line 70: Destination file path 'C:\WINDOWS\System32\drivers' for file 'ndislwf.sys' is not isolated to DIRID 13.

jasonknichel:
@kahashimoto-ms - we saw that your PR moves the FilterDriverParams and FilterAdapterParams keys under the service under the service's Parameters key.  While that is a requirement of 'Windows Drivers', we do not see any changes to NDIS or what might be reading these values to update them to look for these values under a service's parameters key.  Has NDIS, or whatever should be reading these values, been updated to look for these in the service's Parameters key in addition to the legacy location directly under the service's registry state?

jasonknichel (changes requested):
Please see my other comments for things to address in this PR.

kahashimoto-ms:
> @kahashimoto-ms - running 'infverif /v /w /stampinf' on your modified INF still shows a "Windows Driver" violation. Have you taken a look at this violation? Is it possible to address it currently or does the device stack not support moving the file?
> 
> ERROR(1322) in S:\git\testDrv\network\ndis\filter\netlwf.inf, line 70: Destination file path 'C:\WINDOWS\System32\drivers' for file 'ndislwf.sys' is not isolated to DIRID 13.

I have taken a look at this error. I have not yet pushed the changes for this error because there is a problem with installing the driver when DIRID is changed to 13 and the Control Panel Network Installer GUI is used for installation. Specifically, the %13% for the ServiceBinary is not being resolved to the driver store but to "." during installation. I tried debugging this issue but could not quite figure out the root cause. I do believe it could be resolved with an assist from the PnP triage team and have already contacted them for their help. If you prefer to wait until this issue is resolved to push any of the changes I understand.

kahashimoto-ms:
> While that is a requirement of 'Windows Drivers', we do not see any changes to NDIS or what might be reading these values to update them to look for these values under a service's parameters key.

I did think of this before I pushed these changes, but I admit I was not quite sure how to handle this. This registry is not being read by the sample code in the ndislwf project and likely not read anywhere in the entire repo. I believe the purpose of including these writes to the registry is to demonstrate the proper way to do so. I could just comment these lines out. Please advise on the best way to proceed.

PR #811, opened 2022-12-06
Remove invalid INF DeviceInstall32

gerritsMSFT:
Removes from the sample an invalid API call.


jasonknichel (changes requested):
See offline comments

lauragonzalezzz (approved):

PR #825, opened 2023-01-10
Update BthEchoSampleCli and BthEchoSampleSrv to be single architectur…

dnl-q:
…e targeted and the specific build architecture is stamped at the time it is built.


jasonknichel:
The TargetOSVersion for the Models sections for both INFs previously was NT<arch>...1 which scopes this to a ProductType of VER_NT_WORKSTATION (see https://learn.microsoft.com/en-us/windows-hardware/drivers/install/inf-manufacturer-section ).  You removed those ...1 so now the INFs will apply to all product types.  Is that intentional? If you really do support all product types (which drivers usually do), we do support removing product type filtering, but I want to make sure that is intentional.

jasonknichel:
The INFs have:
```
[BthEchoSampleCli_Inst.NT]
CopyFiles=Drivers_Dir

[Drivers_Dir]
BthEchoSampleCli.sys
```
and
```
[BthEchoSampleSrv_Inst.NT]
CopyFiles=Drivers_Dir

[Drivers_Dir]
BthEchoSampleSrv.sys
```

Which aren't specifically called out in the DestinationDirs section so they will use the DefaultDestDir which you updated to DIRID 13 (thanks!).  But that means that these files are no longer copied to the "drivers directory", so the CopyFiles section name of "Drivers_Dir" is misleading.  We should probably rename that section.

jasonknichel:
You changed the DestinationDirs of CopyFiles section [BthEchoSampleCli_Inst_CoInstaller_CopyFiles], but you didn't change the AddReg registering the file copied by the CopyFiles section:
```
[BthEchoSampleCli_Inst_CoInstaller_AddReg]
HKR,,CoInstallers32,0x00010000, "WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll,WdfCoInstaller"
```
However, CoInstallers are (long) deprecated and this WDF coinstaller hasn't been needed in the INF since around Windows 8 and we ignore it (hence why no InfVerif error about it), so you should be able to remove the WDF coinstaller references from your INF.

(Same comments apply to the similar sections/changes in the other INF)

jasonknichel (changes requested):
See comments

dnl-q:
> You changed the DestinationDirs of CopyFiles section [BthEchoSampleCli_Inst_CoInstaller_CopyFiles], but you didn't change the AddReg registering the file copied by the CopyFiles section:
> 
> ```
> [BthEchoSampleCli_Inst_CoInstaller_AddReg]
> HKR,,CoInstallers32,0x00010000, "WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll,WdfCoInstaller"
> ```
> 
> However, CoInstallers are (long) deprecated and this WDF coinstaller hasn't been needed in the INF since around Windows 8 and we ignore it (hence why no InfVerif error about it), so you should be able to remove the WDF coinstaller references from your INF.
> 
> (Same comments apply to the similar sections/changes in the other INF)

Thanks, I removed coinstaller references from both INX files.

dnl-q:
> The INFs have:
> 
> ```
> [BthEchoSampleCli_Inst.NT]
> CopyFiles=Drivers_Dir
> 
> [Drivers_Dir]
> BthEchoSampleCli.sys
> ```
> 
> and
> 
> ```
> [BthEchoSampleSrv_Inst.NT]
> CopyFiles=Drivers_Dir
> 
> [Drivers_Dir]
> BthEchoSampleSrv.sys
> ```
> 
> Which aren't specifically called out in the DestinationDirs section so they will use the DefaultDestDir which you updated to DIRID 13 (thanks!). But that means that these files are no longer copied to the "drivers directory", so the CopyFiles section name of "Drivers_Dir" is misleading. We should probably rename that section.

I renamed Drivers_Dir to DriverStore_Dir. Thanks.

dnl-q:
> The TargetOSVersion for the Models sections for both INFs previously was NT...1 which scopes this to a ProductType of VER_NT_WORKSTATION (see https://learn.microsoft.com/en-us/windows-hardware/drivers/install/inf-manufacturer-section ). You removed those ...1 so now the INFs will apply to all product types. Is that intentional? If you really do support all product types (which drivers usually do), we do support removing product type filtering, but I want to make sure that is intentional.

Recert back to NT$ARCH$...1. Thanks.

jasonknichel:
I see that when you removed the coinstaller references, you also removed:
```
[BthEchoSampleCli_Inst.NT.Wdf]
KmdfService =  BthEchoSampleCli, BthEchoSampleCli_Inst_wdfsect

[BthEchoSampleCli_Inst_wdfsect]
KmdfLibraryVersion = $KMDFVERSION$
```
and
```
[BthEchoSampleSrv_Inst.NT.Wdf]
KmdfService =  BthEchoSampleSrv, BthEchoSampleSrv_Inst_wdfsect

[BthEchoSampleSrv_Inst_wdfsect]
KmdfLibraryVersion = $KMDFVERSION$
```

Those are not tied to the coinstaller and I believe should still be there (the coinstaller would look at them but I believe the new logic that makes the coinstaller unneccessary also looks at those).  Can you please add them back?

jasonknichel:
https://github.com/microsoft/Windows-driver-samples/blob/develop/bluetooth/bthecho/README.md says:
```
Copy KMDF coinstaller (wdfcoinstallerMMmmm.dll, from redist\wdf\ ), BthEchoSampleSrv.Sys, BthEchoSampleSrv.inf and bthsrvinst.exe on a temporary directory on the target machine.
```
and
```
Copy KMDF coinstaller (wdfcoinstallerMMmmm.dll, from redist\wdf\), BthEchoSampleCli.Sys, BthEchoSampleCli.inf and bthecho.exe on a temporary directory on the target machine.
```

With the removal of the coinstaller reference from the INFs, I believe those steps now aren't needed, right?  If so, can you please update the README.md?

dnl-q:
> https://github.com/microsoft/Windows-driver-samples/blob/develop/bluetooth/bthecho/README.md says:
> 
> ```
> Copy KMDF coinstaller (wdfcoinstallerMMmmm.dll, from redist\wdf\ ), BthEchoSampleSrv.Sys, BthEchoSampleSrv.inf and bthsrvinst.exe on a temporary directory on the target machine.
> ```
> 
> and
> 
> ```
> Copy KMDF coinstaller (wdfcoinstallerMMmmm.dll, from redist\wdf\), BthEchoSampleCli.Sys, BthEchoSampleCli.inf and bthecho.exe on a temporary directory on the target machine.
> ```
> 
> With the removal of the coinstaller reference from the INFs, I believe those steps now aren't needed, right? If so, can you please update the README.md?

Done, thanks.

dnl-q:
> I see that when you removed the coinstaller references, you also removed:
> 
> ```
> [BthEchoSampleCli_Inst.NT.Wdf]
> KmdfService =  BthEchoSampleCli, BthEchoSampleCli_Inst_wdfsect
> 
> [BthEchoSampleCli_Inst_wdfsect]
> KmdfLibraryVersion = $KMDFVERSION$
> ```
> 
> and
> 
> ```
> [BthEchoSampleSrv_Inst.NT.Wdf]
> KmdfService =  BthEchoSampleSrv, BthEchoSampleSrv_Inst_wdfsect
> 
> [BthEchoSampleSrv_Inst_wdfsect]
> KmdfLibraryVersion = $KMDFVERSION$
> ```
> 
> Those are not tied to the coinstaller and I believe should still be there (the coinstaller would look at them but I believe the new logic that makes the coinstaller unneccessary also looks at those). Can you please add them back?

Done, thanks.

jasonknichel (approved):

PR #844, opened 2023-01-30
Build scripts update

NeoAdonis:
* Updates script to build samples concurrently. For GitHub Actions, it saves about 60% of usage time.
* Provides documentation to some of the scripts for ease of usage.
* Replaces the usage of "Project" to "Sample" to avoid confusion with Visual Studio projects.
* Other miscellaneous fixes, such as fixing an issue where files with spaces in their names were not recognized correctly while running the build workflow when pull requests were created.


NeoAdonis:
CodeQL Analysis is failing due to issues with lack of disk space, but otherwise seems to run the same compared to before the script changes. Based on this and leveraging the results from the other pipelines, I'll merge this change.

PR #854, opened 2023-02-06
IoStatus.Information should not be set on cleanup/close. Appears to b…

jeffbromberger:
IoStatus.Information should not be set on cleanup/close. Appears to be copy and paste from Create code.


rodwiddowson:
I’m AFI ATM, BUT I’ll observe that NTFS does set information in the CLEANUP path under circumstances I forget..

Also it’s probably a good idea to zero out information anyway.

(Ps. Hi Jeff!)

jeffbromberger:
Hi Rod!

Setting it to zero sounds good.  

I'm quite interested in the notion that NTFS would set it to something purposeful.  If this is indeed a field to be set then there is a separate documentation bug for the IRP_MJ_CLEANUP page.  Fwiw, the CDFS sample doesn't appear to set it to anything (other than zeroing it on error in CdCompleteRequest).  The WRK (admittedly super old) doesn't appear to even check the status, much less the information, on IRP_MJ_CLEANUP.

JakobL-MSFT:

JakobL-MSFT:

PR #880, opened 2023-02-14
Fix driver violations for energy team owned drivers

mvillalpando06:


simbatt/func/simbatt.vcxproj.Filters
```diff
@@ -41,6 +41,12 @@
     <ClInclude Include="*.h;*.hpp;*.hxx;*.hm;*.inl;*.xsd">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="*.h;*.hpp;*.hxx;*.hm;*.inl;*.xsd">
```
jasonknichel, line 44: Should the same ClInclude be repeated multiple times?

simbatt/func/simbatt.vcxproj.Filters
```diff
@@ -49,5 +55,8 @@
     <Inf Include="*.inx">
       <Filter>Driver Files</Filter>
     </Inf>
+    <Inf Include="*.inx">
```
jasonknichel, line 58: Should the same "Inf Include=" be repeated multiple times?

jasonknichel (changes requested):

thermal/simsensor/simsensor.vcxproj
```diff
@@ -30,31 +30,31 @@
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <TargetVersion>Windows10</TargetVersion>
     <UseDebugLibraries>False</UseDebugLibraries>
-    <DriverTargetPlatform>Universal</DriverTargetPlatform>
+    <DriverTargetPlatform>Windows Driver</DriverTargetPlatform>
```
jasonknichel, line 33: For simsensor.inf, you will have to leave it as Universal for now.  I'll reach out to you offline to discuss

simbatt/func/simbatt.vcxproj.Filters
mvillalpando06: these get added when the project is built. Tried removing them and rebuilding and got added again.

simbatt/func/simbatt.inx
jasonknichel, line 1: Files using DIRID 13 need to use TargetOSVersion targeting so that they only install on builds 16299 and later (the first release we support DIRID 13 on). See https://learn.microsoft.com/en-us/windows-hardware/drivers/install/combining-platform-extensions-with-operating-system-versions for more information on TargetOSVersion

thermal/thermalclient/simtc.inf
jasonknichel, line 1: Files using DIRID 13 need to use TargetOSVersion targeting so that they only install on builds 16299 and later (the first release we support DIRID 13 on). See https://learn.microsoft.com/en-us/windows-hardware/drivers/install/combining-platform-extensions-with-operating-system-versions for more information on TargetOSVersion

jasonknichel (changes requested):

jasonknichel (approved):

PR #881, opened 2023-02-14
Fix logging when there are errors

NeoAdonis:
A quick fix to a bug where errors were not shown even if builds would actually fail. This change also introduces a simple way to see details about the failed samples in the meantime while a better solution to make logs available is designed.

### DIFF
--- a/Build-SampleSet.ps1
+++ b/Build-SampleSet.ps1
@@ -167,9 +167,13 @@ $sw.Stop()
 if ($jresult.FailSet.Count -gt 0) {
     Write-Output "Some combinations were built with errors:"
     foreach ($failedSample in $jresult.FailSet) {
-        Write-Output "== $failedSample =="
         $failedSample -match "^(.*) (\w*)\|(\w*)$" | Out-Null
-        Get-Content "$LogFilesDirectory\$($Matches[1]).$($Matches[2]).$($Matches[3]).err" | Write-Output
+        $failName = $Matches[1]
+        $failConfiguration = $Matches[2]
+        $failPlatform = $Matches[3]
+        Write-Output "Build errors in Sample $failName; Configuration: $failConfiguration; Platform: $failPlatform {"
+        Get-Content "$LogFilesDirectory\$failName.$failConfiguration.$failPlatform.err" | Write-Output
+        Write-Output "} $failedSample"
     }
     Write-Error "Some combinations were built with errors."
 }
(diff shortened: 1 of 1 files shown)


Build-SampleSet.ps1
```diff
@@ -162,10 +164,12 @@ $SampleSet.GetEnumerator() | ForEach-Object -ThrottleLimit $ThrottleLimit -Paral
 
 $sw.Stop()
 
-if ($failSet.Count -gt 0) {
+if ($jresult.FailSet.Count -gt 0) {
     Write-Output "Some combinations were built with errors:"
-    foreach ($failedSample in $failSet) {
-        Write-Output "$failedSample"
+    foreach ($failedSample in $jresult.FailSet) {
+        Write-Output "== $failedSample =="
```
JakobL-MSFT, line 170: Hi.  Can I ask that the "Write-Output" follows same style as what we use above with curly braces.  So something like this:
---
Failure in Sample: general.tracing.evntdrv; Configuration: Debug; Platform: x64 {
...
Failure in Sample: general.tracing.evntdrv; Configuration: Debug; Platform: x64 }
---

JakobL-MSFT:
See above for a comment about output format.

Secondly see the Code Analysis CI gate failing.  I am surprised about that.  Perhaps this is from an earlier commit to develop?

NeoAdonis:
I noticed the error on the CodeQL Analysis workflow. It seems that the issue here is quite different, caused by the submodules not being populated correctly on repo checkout. Silver lining: it means that this change is working as expected, as the errors are now being shown properly.

I'll fix this issue in a separate PR.

Build-SampleSet.ps1
NeoAdonis: Fixed format. Output message is now similar to:

```
Build errors in Sample: general.tracing.evntdrv; Configuration: Debug; Platform: x64 {
(Actual error output here)
} general.tracing.evntdrv Debug|x64
```

JakobL-MSFT:
Adonais:

Code Analysis CI gate failing: This indeed goes all the way back to this old PR: https://github.com/microsoft/Windows-driver-samples/pull/863 and your change indeed just makes it visible. Great.

See above for a comment about output format.

JakobL-MSFT (changes requested):
Adonais:

Code Analysis CI gate failing: This indeed goes all the way back to this old PR: https://github.com/microsoft/Windows-driver-samples/pull/863 and your change indeed just makes it visible. Great.

See above for an unrelated comment about output format.

NeoAdonis:
Jakob;

I have already resolved your comment about output format on the latest commit. Is there a different issue that I'm missing?

JakobL-MSFT (approved):

lauragonzalezzz (approved):

PR #882, opened 2023-02-14
Adding the AudioCodec ACX sample driver to Github.

judzmura:
This PR adds the first ACX sample audio driver to Github. This driver will be used as a reference point for partners who are trying to write ACX audio drivers of their own. The AudioCodec folder holds the files that are just for this first sample audio driver. The Common/Inc/Shared folders contain files that are used in this driver but can be used in other sample ACX audio drivers as well. I have two other such drivers that will eventually be published on Github but still need some validation. I will also be adding a Readme for this sample driver and will be adding information about it to the public ACX documentation so there will be instruction on how to build it, etc.
I built the sample using the latest public NI WDK (first build the common library, then the AudioCodec driver which is dependent on that library) and made sure it installs and works property and HLK is passing. Also confirmed that the driver is Windows Drivers compliant.


audio/Acx/Samples/AudioCodec/Driver/ReadMe.txt
```diff
@@ -0,0 +1,42 @@
+========================================================================
```
jasonknichel, line 1: Maybe make this a .md file with markdown formatting?

PR #886, opened 2023-02-19
Update Building-Locally.md to remove "`" 

ahidaka:
Removed "`" which causes problem. Then it becomes the following clean line: winget install --id Git.Git --source winget


ahidaka:
Thanks! It moved to #901 for develop branch.

PR #888, opened 2023-02-20
Fix echoum.inx

ahidaka:
This INX file has some problems that cause errors in recent WDK environments, so I fix them. I have confirmed that the umdf2echo driver can be successfully built and works with this fix in a recent WDK environment. This is always easily verified by anyone.


jasonknichel:
Thank you for the sample update @ahidaka!  We are asking that changes primarily happen in the 'develop' branch of the repo.  Could you rebase your change to that branch?  Also, it looks like you added a BOM to the INF file and it seems like the INF file (well .inx file in this case) may be UTF-8.  Windows requires that INFs be ANSI or UTF-16 LE with UTF-16 LE preferred.  Can you please make sure the INF file is UTF-16 LE?

ahidaka:
Thanks! It moved to #901 for develop branch.

PR #901, opened 2023-02-25
Fix Building-Locally.md and echoum.inx

ahidaka:
I send a new PR to the develop branch instead of #886 and #888.

#886 simply solves the problem that occurs when copy-pasting commands.
#888 allows to build objects using Visual Studio 2002.


wm1:
Actually, I'd prefer to **NOT** change echoum.inx. Add @NeoAdonis for comment.

Some people use the WDK sample as-is without any modification _in production_. That is bad, because if two people are doing this, their driver packages will conflict with each other's.

WDK tries to report such error early on. To do so, the samples are _intentionally_ using the forbidden class name "Sample", forbidden service name like "echo". The sample now fails to build immediately. This forces the user to make _some_ modification to the sample before publishing it as their own code.

PR #910, opened 2023-03-03
Opt many samples in for Universal validation

michbern-ms:
Many of our driver samples build cleanly agains the UniversalDDIs.xml
API list.  Opting these in as Universal helps protect their API surface
against accidentally picking up more Desktop-only APIs.


michbern-ms:
This also includes a commit to shift avstream/sampledevicemft sample to use mfcore.dll rather than mf.dll - this works better on some of our SKUs that have mfplat.dll and mfcore.dll but lack the larger mf.dll.

jasonknichel (approved):

PR #935, opened 2023-03-17
[avstream/sampledevicemft] Fix possible deadlock while changes output stream state

jammgit:
prerequisites:
There are multiple output streams.
BUG:
In the CMultipinMft::ProcessOutput function, when the first stream object is obtained and executed to spOpin->ProcessOutput, SetOutputStreamState may be executed at the same time to change the stream state. The latter acquires the lock and begins to wait for the Input stream state change. At this point, the CMultipinMft::ProcessOutput function starts to process the second stream, and cannot acquire the lock. Finally, the DTM deadlock...


jammgit:

PR #973, opened 2023-04-11
Ndis mux inf violations

kahashimoto-ms:
Fix INF violations. Fix Prefast warnings. Fix encoding on INF file. Add the correct build number needed to install using DIRID 13 for packages that have a notify object DLL.


network/ndis/mux/notifyob/adapter.cpp
```diff
@@ -225,9 +229,14 @@ HRESULT CMuxPhysicalAdapter::LoadConfiguration (VOID)
...
                         if ( lpMiniportGuid != NULL ) {
 
-                            CLSIDFromString( lpMiniportGuid,
+                            HRESULT hrResult = CLSIDFromString( lpMiniportGuid,
                                              &guidMiniport );
-         
+
+                            if (hrResult != NOERROR)
+                            {
+                                lResult = E_INVALIDARG;
```
jasonknichel, line 237: lResult is used above to capture returns fro Reg* APIs which return Win32 errors, but you are assigning a HRESULT into it here.  You shouldn't mix HRESULT and Win32 error like this.  Should this be lResult = ERROR_INVALID_PARAMETER; or lResult = ERROR_INVALID_DATA; instead?

network/ndis/mux/notifyob/adapter.cpp
```diff
@@ -146,9 +146,13 @@ HRESULT CMuxPhysicalAdapter::LoadConfiguration (VOID)
     // device IDs of the virtual miniports are stored.
     //
 
-    StringFromGUID2( m_guidAdapter,
+    int numChars = StringFromGUID2( m_guidAdapter,
                     szAdapterGuid,
                     MAX_PATH+1 );
+    if (numChars == 0)
+    {
+        return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 154: This function returns an HRESULT but ERROR_BUFFER_OVERFLOW is a Win32 error.  Should this be HRESULT_FROM_WIN32(ERROR_BUFFER_OVERFLOW); instead?

network/ndis/mux/notifyob/common.cpp
```diff
@@ -336,10 +337,15 @@ HRESULT HrFindInstance (INetCfg *pnc,
 
     if ( hr == S_OK ) {
 
-        StringFromGUID2( guidInstance,
+        numChars = StringFromGUID2( guidInstance,
                       szGuid,
                       MAX_PATH+1 );
 
+        if (numChars == 0)
+        {
```
jasonknichel, line 345: This and other added code doesn't match the existing curly brace style of this file

network/ndis/mux/notifyob/common.cpp
```diff
@@ -353,10 +359,16 @@ HRESULT HrFindInstance (INetCfg *pnc,
...
             if ( hr == S_OK ) {
 
-                StringFromGUID2( guid,
+                numChars = StringFromGUID2( guid,
                                 szGuid,
                                 MAX_PATH+1 );
 
+                if (numChars == 0)
+                {
+                    return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 368: This function returns an HRESULT but ERROR_BUFFER_OVERFLOW is a Win32 error. Should this be HRESULT_FROM_WIN32(ERROR_BUFFER_OVERFLOW); instead?

network/ndis/mux/notifyob/common.cpp
```diff
@@ -336,10 +337,15 @@ HRESULT HrFindInstance (INetCfg *pnc,
...
     if ( hr == S_OK ) {
 
-        StringFromGUID2( guidInstance,
+        numChars = StringFromGUID2( guidInstance,
                       szGuid,
                       MAX_PATH+1 );
 
+        if (numChars == 0)
+        {
+            return ERROR_BUFFER_OVERFLOW;
```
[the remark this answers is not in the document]
jasonknichel: Also, is this early return leaking pencc since ReleaseObj( pencc ); is being skipped?

network/ndis/mux/notifyob/common.cpp
jasonknichel: Also, is this early return leaking pencc since ReleaseObj( pencc ); is being skipped?

network/ndis/mux/notifyob/notify.cpp
```diff
@@ -1232,9 +1229,14 @@ HRESULT CMuxNotify::HrLoadAdapterConfiguration (VOID)
                 // Subkeys are actually a guid/bindname of the adapters.
                 //
                 szAdapterGuid[MAX_PATH]='\0';
-                CLSIDFromString( szAdapterGuid,
+                HRESULT hrResult = CLSIDFromString( szAdapterGuid,
                                  &guidAdapter );
 
+                if (hrResult != NOERROR)
+                {
```
jasonknichel, line 1236: not matching existing curly brace style

network/ndis/mux/notifyob/notify.cpp
```diff
@@ -1232,9 +1229,14 @@ HRESULT CMuxNotify::HrLoadAdapterConfiguration (VOID)
...
+                    lResult = E_INVALIDARG;
```
jasonknichel, line 1237: You are assigning an HRESULT to lResult which this function uses for Win32 errors.

network/ndis/mux/notifyob/virtual.cpp
```diff
@@ -330,17 +330,22 @@ HRESULT CMuxVirtualMiniport::ApplyRegistryChanges(ConfigAction eApplyAction)
...
 
         case eActAdd:         // Virtual miniport added.
 
-            StringFromGUID2( m_guidAdapter,
+            numChars = StringFromGUID2( m_guidAdapter,
                              szAdapterGuid,
                              MAX_PATH+1 );
 
+            if (numChars == 0)
+            {
```
jasonknichel, line 345: not matching the brace style of existing code

network/ndis/mux/notifyob/virtual.cpp
```diff
@@ -330,17 +330,22 @@ HRESULT CMuxVirtualMiniport::ApplyRegistryChanges(ConfigAction eApplyAction)
...
+                return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 346: you are returning a win32 error here, but this function should return an HRESULT.

network/ndis/mux/notifyob/virtual.cpp
```diff
@@ -361,10 +366,15 @@ HRESULT CMuxVirtualMiniport::ApplyRegistryChanges(ConfigAction eApplyAction)
...
             if ( lResult == ERROR_SUCCESS ) {
 
-                StringFromGUID2( m_guidMiniport,
+                numChars = StringFromGUID2( m_guidMiniport,
                                  szMiniportGuid,
                                  MAX_PATH+1 );
 
+                if (numChars == 0)
+                {
+                    return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 375: you are returning a win32 error here, but this function should return an HRESULT.

jasonknichel: this early return is leaking szAdapterGuidKey

network/ndis/mux/notifyob/virtual.cpp
```diff
@@ -411,10 +421,15 @@ HRESULT CMuxVirtualMiniport::ApplyRegistryChanges(ConfigAction eApplyAction)
...
         case eActRemove:                  // Virtual miniport removed.
 
-            StringFromGUID2( m_guidAdapter,
+            numChars = StringFromGUID2( m_guidAdapter,
                              szAdapterGuid,
                              MAX_PATH+1 );
 
+            if (numChars == 0)
+            {
+                return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 430: you are returning a win32 error here, but this function should return an HRESULT.

network/ndis/mux/notifyob/virtual.cpp
```diff
@@ -434,10 +449,15 @@ HRESULT CMuxVirtualMiniport::ApplyRegistryChanges(ConfigAction eApplyAction)
...
             if ( lResult == ERROR_SUCCESS ) {
 
-                StringFromGUID2( m_guidMiniport,
+                numChars = StringFromGUID2( m_guidMiniport,
                                  szMiniportGuid,
                                  MAX_PATH+1 );
 
+                if (numChars == 0)
+                {
+                    return ERROR_BUFFER_OVERFLOW;
```
jasonknichel, line 458: you are returning a win32 error here, but this function should return an HRESULT.

jasonknichel: this early return is leaking szAdapterGuidKey

network/ndis/mux/driver/60/mux_mp.inf
jasonknichel, line 1: This is in class Net, so do the comments about the Network Installation UI and the build number filtering still apply or was that just for the NetTrans and/or NetService/NetClient ones you modified earlier?

jasonknichel, line 1: mux_mp.inf adds a service that uses mux.sys, but it doesn't payload mux.sys.  You need a [SourceDisksNames], [SourceDisksFiles], [DestinationDirs], and CopyFiles for it.

network/ndis/mux/driver/60/muxp.inf
jasonknichel, line 1: You have some "CHECK THIS" comments still in there:

MUXP_NC.CopyFiles.DLL  = 11 ;CHECK THIS
MUXP_NC.CopyFiles.Sys  = 12    ;CHECK THIS

jasonknichel (changes requested):

network/ndis/mux/driver/60/mux_mp.inf
kahashimoto-ms: This is an intermediate driver which require two INF files. One of the INF files (muxp.inf) defines the installation for the protocol lower edge. The other INF file (mux_mp.inf) defines the installation for the virtual miniport upper edge. The protocol INF, (muxp.inf) is the primary INF file. After the protocol is installed the miniport upper edge is installed. The protocol INF installs a DLL, which can only be properly installed with OS build number 25341 or latter. The miniport INF does not install a DLL so the build number of 25341 is not necessary. However, I thought it would make more sense if the two INF files used the same DIRID during installation. If the build number was set to a lower number in the miniport driver INF, you could end up in a situation where the protocol was installed using DIRID 12 and the miniport was installed using DIRID 13. I am not sure that this is a bad thing, I just thought it could cause some confusion. That is why the build number of 25341 is used in both INF files.

kahashimoto-ms: Should I change the comment to explain that?

kahashimoto-ms: I edited the comment to explain why the build number is 25341 in this INF file as well.

jasonknichel (changes requested):

jasonknichel (approved):

PR #977, opened 2023-04-18
[general/simplemediasource] Fix architecture in solution

NeoAdonis:
When building the solution, Debug|ARM64 was treated as Debug|Win32 in the simplemediasource project. This has been fixed and the exclusion related to this error has been removed.


DavidS-msft (approved):
LGTM

PR #995, opened 2023-06-14
Firefly driver: Postpone variable definitions until first use

forderud:
Motivation:
* Improves readability and maintainability.
* Have been supported since C99.
* _Can_ potentially reduce stack size usage when variable usage is confined to a local scope. Example: `usage` and `usageLength` variables in hid/firefly/driver/vfeature.c. The compiler is probably already optimizing this though.

Downside:
* It's a bit more cumbersome to manually determine the stack size of a function when all local variables are no longer listed together. However, tools can still do this effectively.

### DIFF
--- a/hid/firefly/driver/device.c
+++ b/hid/firefly/driver/device.c
@@ -53,13 +53,6 @@ Return Value:
 
 --*/    
 {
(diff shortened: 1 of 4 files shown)


JakobL-MSFT:

PR #1013, opened 2023-08-01
Fsfilters isolation updates

zlockard:


cgallred (changes requested):
We do not want to put version checks in public samples; checking the OS version is not the best way to determine whether a feature exists. Besides that, the samples won't build on a version of Windows that doesn't have `IoOpenDriverRegistryKey` exported anyway, so the version check doesn't help.

Instead, you should use the [MmGetSystemRoutineAddress](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-mmgetsystemroutineaddress) function to probe for the existence of `IoOpenDriverRegistryKey`, and then use it if it exists.

zlockard:
> We do not want to put version checks in public samples; checking the OS version is not the best way to determine whether a feature exists. Besides that, the samples won't build on a version of Windows that doesn't have `IoOpenDriverRegistryKey` exported anyway, so the version check doesn't help.
> 
> Instead, you should use the [MmGetSystemRoutineAddress](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-mmgetsystemroutineaddress) function to probe for the existence of `IoOpenDriverRegistryKey`, and then use it if it exists.

Yeah if you look at the latest cancelsafe in this PR it uses MmGetSystemRoutineAddress, we're still looking at what code paradigm we want for this

cgallred (approved):

PR #1017, opened 2023-08-08
Remove Print Driver Samples

alexke-dev:
This change includes the following:
- Removes print driver samples.
- Adds a README.md to redirect to the Print Support App (PSA) guide.


print/README.md
```diff
@@ -0,0 +1,5 @@
+# Printer Device Development with Print Support Apps (PSA)
+
+We recommend that you use Microsoft's IPP inbox class driver, along with Print Support Apps (PSA), to customize the print experience in Windows 10 and 11 for printer device development.
+
+For more information, see [Print support app design guide](https://learn.microsoft.com/en-us/windows-hardware/drivers/devapps/print-support-app-design-guide).
```
jasonknichel, line 5: You should remove the "en-us" from the learn.microsoft.com link so the page will load in the default language of the user clicking the link.

alexke-dev: Fixed

VladiStep:
"Thank" you (or someone who made the decision) very much 😠
Good thing that GitHub preserves all the files.

echowangbj:
where can I get the sample code?

PR #1019, opened 2023-08-09
Added pnputil equivalents to devcon help text output

emilyhobby:
Add pnputil equivalents to devcon help text output. This output supports the ongoing efforts to migrate from devcon to pnputil.

### DIFF
--- a/setup/devcon/msg.mc
+++ b/setup/devcon/msg.mc
@@ -25,6 +25,10 @@ For more information, type: %1 help %2
 ;//
 MessageId=60100 SymbolicName=MSG_HELP_LONG
 Language=English
+------------------------------------------------------------------------------
+|           Devcon is a code sample intended for learning purposes           |
+|       Use PnpUtil for optimal functionality/compatiblity with this OS      |
+------------------------------------------------------------------------------
 Device Console Help:
 %1 [-r] [-m:\\<machine>] <command> [<arg>...]
 -r           Reboots the system only when a restart or reboot is required.
@@ -54,6 +58,8 @@ Lists all device setup classes. Valid on local and remote computers.
 <machine>    Specifies a remote computer.
 Class entries have the format <name>: <descr>
 where <name> is the class name and <descr> is the class description.
+
+Recommended Equivalent: pnputil /enum-classes
 .
 MessageId=60201 SymbolicName=MSG_CLASSES_SHORT
 Language=English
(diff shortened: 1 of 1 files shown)


setup/devcon/msg.mc
```diff
@@ -25,6 +25,10 @@ For more information, type: %1 help %2
 ;//
 MessageId=60100 SymbolicName=MSG_HELP_LONG
 Language=English
+------------------------------------------------------------------------------
+|           Devcon is a code sample intended for learning purposes           |
```
jasonknichel, line 29:
```suggestion
|           Devcon is a code sample intended for learning purposes.          |
```

setup/devcon/msg.mc
```diff
@@ -598,6 +628,8 @@ Installs the specified device manually. Valid only on the local computer.
 <inf>        Specifies an INF file with installation information for the device.
 <hwid>       Specifies a hardware ID for the device.
 -r           Reboots the system only when a restart or reboot is required.
+
+Recommended Equivalent: pnputil /add-driver <inf> /install
```
jasonknichel, line 632: Should this also mention devgen?

jasonknichel: also, this recommendation of "pnputil /add-driver <inf> /install" should be added to devcon update text below.

setup/devcon/msg.mc
```diff
@@ -827,6 +869,8 @@ Examples of <id>:
  @ISAPNP\*\*    - Instance ID with wildcards  (@ prefixes instance ID)
  '*PNP0501      - Hardware ID with apostrophe (' prefixes literal match - matches exactly as typed,
                                                including the asterisk.)
+
+Recommended Equivalent: pnputil /remove-device /deviceid <id>
```
jasonknichel, line 873: why different recommendation between devcon remove and devcon removeall?

emilyhobby: pnputil /remove-device /deviceid <id> allows users to remove all devices that match on a specific hardware ID, while pnputil /remove-device <instance id> removes a single device. /deviceid more closely aligns with removeall

jasonknichel: You can remove multiple devices with 'devcon remove' also. Aren't 'devcon remove' and 'devcon removeall' the same other than 'devcon remove' can only work on present devices and 'devcon removeall' can also include non-present devices (see https://github.com/microsoft/Windows-driver-samples/commit/389bdb217d54112d96b40bc967f86b17a605bed6 )?  If that is correct, then from the pnputil command perspective, the pnputil commands for both should be the same (since pnputil /remove-device would just work on all devices, present or not)

jasonknichel (approved):

Issue #1025, opened 2023-09-03
Reserved service name errors when building Toaster KMDF samples


forderud:
I'm experiencing the following errors when attempting to build the [general\toaster\toastDrv](https://github.com/microsoft/Windows-driver-samples/tree/main/general/toaster/toastDrv) KMDF projects with Visual Studio 2022:
```
>------ Build started: Project: dynambus, Configuration: Debug x64 ------
>Building 'dynambus' with toolset 'WindowsKernelModeDriver10.0' and the 'Universal' target platform.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\dynamic\dynambus.inx(64-64): error 1293: Service name 'dynambus' is reserved.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\dynamic\dynambus.inx(75-75): error 1293: Service name 'dynambus' is reserved.

>------ Build started: Project: StatBus, Configuration: Debug x64 ------
>Building 'StatBus' with toolset 'WindowsKernelModeDriver10.0' and the 'Universal' target platform.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\static\statbus.inx(64-64): error 1293: Service name 'statbus' is reserved.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\static\statbus.inx(75-75): error 1293: Service name 'statbus' is reserved.

>------ Build started: Project: filter (Kmdf\Filter\Generic\filter), Configuration: Debug x64 ------
>Building 'filter' with toolset 'WindowsKernelModeDriver10.0' and the 'Universal' target platform.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\filter\filter.inx(76-76): error 1293: Service name 'ToasterFilter' is reserved.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\filter\filter.inx(77-77): error 1293: Service name 'wdffeatured' is reserved.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\filter\filter.inx(96-96): error 1293: Service name 'ToasterFilter' is reserved.

>------ Build started: Project: filter (Kmdf\Filter\Sideband\filter), Configuration: Debug x64 ------
>Building 'filter' with toolset 'WindowsKernelModeDriver10.0' and the 'Universal' target platform.
[...]

PR #1027, opened 2023-09-03
Rename "dynambus" to "dynamicbus"

forderud:
Proposal for partial fix to #1025. Done to avoid the following build error:
```
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\dynamic\dynambus.inx(64-64): error 1293: Service name 'dynambus' is reserved.
>C:\Windows-driver-samples\general\toaster\toastDrv\kmdf\bus\dynamic\dynambus.inx(75-75): error 1293: Service name 'dynambus' is reserved.
```


JakobL-MSFT:

wm1:
The compile failure is intentional. The idea is to prevent somebody from copy the sample without updating service name etc.  Thus, reject the PR.

forderud:
> The compile failure is intentional. The idea is to prevent somebody from copy the sample without updating service name etc. Thus, reject the PR.

Understood. Still, it's confusing with sample projects that doesn't build as-is, especially for new driver developers. Would it be possible to instead document this on the project front-page with clear instructions on how to make the projects build?

PR #1040, opened 2023-10-06
Remove .exe from the INF file

zlockard:


jasonknichel (approved):

cgallred:
@jasonknichel how can I look at the changes? GitHub thinks these .inf files are binary, I think because they are UTF16, so it won't show me the diffs.

jasonknichel:
> @jasonknichel how can I look at the changes? GitHub thinks these .inf files are binary, I think because they are UTF16, so it won't show me the diffs.

@cgallred - yea, git and github don't like UTF-16 files and considers them binary. :(  It only does a proper text diff for ASCII and UTF-8.  I diff INF files (which PNP requires to be ASCII or UTF-16 LE) by doing the diff locally in my local clone of the sample repo. For this change, I saw that the commit ID was b0dad9afddee741cbb5422fa8bacdc6a545f4576 so in my local clone, I did this to view the changes:

git difftool -d b0dad9afddee741cbb5422fa8bacdc6a545f4576~1 b0dad9afddee741cbb5422fa8bacdc6a545f4576

PR #1049, opened 2023-11-07
Add new flag and exclusion for new InfVerif changes

NeoAdonis:
Temporary change to fix InfVerif build errors when using the build script with EEAP WDK and samples in development.


JakobL-MSFT (approved):
I agree with this change for today.  

Feature creep (feel free to ignore): I'd prefer if we can have an EOL at end of exclusion.csv .

JakobL-MSFT:
I agree with this change for today.

Feature creep (feel free to ignore): I'd prefer if we can have an EOL at end of exclusion.csv .

PR #1059, opened 2023-12-03
Revert "Delete video/pixlib directory" and instead move it into an archive folder

zhuman:
Reverts microsoft/Windows-driver-samples#1058 and instead places it into an archive folder to preserve for historical purposes.


clandrew:
Looks great!

PR #1062, opened 2023-12-07
simbatt: Remove separate KMDF co-installer INX

forderud:
This INF file should according to [Specifying the KMDF Co-installer in an INF File](https://learn.microsoft.com/en-us/windows-hardware/drivers/wdf/installing-the-framework-s-co-installer) no longer be needed if targeting Windows 10 or newer. Therefore, proposing to delete the file to clean up since the remaining `simbatt.inx` should suffice.


forderud:
Closing PR, since `simbatt-wdfcoinstall.inx` already appear to have been deleted on the "develop" branch in afa69cdeeb4644e8cd9a1a567dc91a7f29947a5b.

PR #1082, opened 2024-01-21
This PR disables infverif checks until samples are ready

JakobL-MSFT:


JakobL-MSFT:
This PR won't work as the NI EWDK does not support /sw*

JakobL-MSFT:
This PR won't work as the NI EWDK does not support /sw*

PR #1092, opened 2024-01-29
Add arch variable to simsensor and simtc infs

mvillalpando06:
Add arch variable to inf files to use specific architecture when building driver solution.


jasonknichel:
you need to update the Models section names too.  i.e. you need to collapse these into a single section that uses $ARCH$ 👍 
image

jasonknichel (approved):

JakobL-MSFT:
Hi @mvillalpando06 . I would like to ask for an entry added to https://github.com/microsoft/Windows-driver-samples/blob/develop/.github/CODEOWNERS as part of this PR.  Contact @NeoAdonis or I offline for further details.

JakobL-MSFT (changes requested):
I would like to ask for an entry added to https://github.com/microsoft/Windows-driver-samples/blob/develop/.github/CODEOWNERS as part of this PR. Contact @NeoAdonis or I offline for further details.

JakobL-MSFT (approved):

PR #1117, opened 2024-02-16
Add explicit flag for arm64 for ehsc and remove incremental link

prashantchahar:
Add explicit flag for arm64 for ehsc and remove incremental link as ehsc is not passsed for ARM64 VS and looks like ltcg is not supported as well.

### DIFF
--- a/video/IndirectDisplay/IddSampleApp/IddSampleApp.vcxproj
+++ b/video/IndirectDisplay/IddSampleApp/IddSampleApp.vcxproj
@@ -141,7 +141,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
-      <AdditionalOptions>/EHsc</AdditionalOptions>
+      <ExceptionHandling>Sync</ExceptionHandling>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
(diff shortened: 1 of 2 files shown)


video/IndirectDisplay/IddSampleApp/IddSampleApp.vcxproj
```diff
@@ -166,6 +141,7 @@
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
+      <AdditionalOptions>/EHsc</AdditionalOptions>
```
NeoAdonis, line 144: Can you check if you can use the following property instead?

```
<ClCompile>
    <ExceptionHandling>Sync</ExceptionHandling>
</ClCompile>
```

NeoAdonis (commented):
lgtm. I left a comment but let me know if that approach isn't useful.

video/IndirectDisplay/IddSampleApp/IddSampleApp.vcxproj
JakobL-MSFT: Hi @NeoAdonis .  Can you explain why you are making this suggestion?  (I have a guess, but would like to learn from you)

NeoAdonis: So that it can be controlled programmatically and also so that it can be configured from the VS GUI:

image

JakobL-MSFT: @prashantchahar to look at above please.

PR #1124, opened 2024-02-21
Fix TrEE Sample

JakobL-MSFT:
The TrEE sample is literally the last sample on our .\exclusion.csv list.  Let's get it to build!!

This fix does narrow fixes to achieve exactly that and at same time adds arm64 as a target OS.

More specifically:
* This PR removes TrEE from the .\exclusion.csv list.
* This PR adds arm64 target support to vcxproj and sln file.
* This PR avoid including non-existing header file and instead adds typedef directly (see beneath).
* This PR includes project header files with "..\inc\foo.h" rather than <foo.h>.  I am not quite sure what the best way to write this is.  But at least this works :-) .
* This PR fixes a requirement to not use ExAllocatePoolWithTag

How tested:
* Built for all four flavors in:
* WDK 22621
* EWDK.ni_release_svc_prod1.22621.2428
* EWDK.ge_release.26063.1

Note: There is no suitable header file currently available that defines CHAR16:
```
C:\Program Files (x86)\Windows Kits\10\Include>findstr /s /m CHAR16 *
10.0.22621.0\km\treevariableservice.h
10.0.22621.0\um\icu.h
10.0.22621.0\um\icucommon.h
10.0.22621.0\um\mi.h
10.0.22621.0\um\WbemCli.h
10.0.22621.0\um\WbemCli.Idl
```

### DIFF
--- a/TrEE/Miniport/SampleMiniport.c
+++ b/TrEE/Miniport/SampleMiniport.c
@@ -22,8 +22,13 @@ Environment:
 #include <wdmguid.h>
 #include <ntstrsafe.h>
 #include <TrustedRuntimeClx.h>
-//#include <ntefi.h>
-typedef UINT16          CHAR16;
+
+//
+// Header file <TrEEVariableService.h> requires prior definition of a CHAR16.
+// For now we need to define this in sample.
+//
+typedef UINT16 CHAR16;
+
 #include <TrEEVariableService.h>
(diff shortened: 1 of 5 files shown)


JakobL-MSFT:
I am going to complete this PR as I would like it my build for tonight and clearly we are better off with this change than without.  More feedback is welcome after completion.

PR #1134, opened 2024-03-04
Update Code-Scanning.yml jobs to use Node.js 20

jacob-ronstadt:
actions/checkout to v4 and microsoft/setup-msbuild to v2

### DIFF
--- a/.github/workflows/Code-Scanning.yml
+++ b/.github/workflows/Code-Scanning.yml
@@ -33,7 +33,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: 'recursive'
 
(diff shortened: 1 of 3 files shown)


JakobL-MSFT:
I may be creating more work here, but we should make sure CodeQL workflow is aligned as close as reasonably possible with https://github.com/microsoft/Windows-driver-samples/blob/main/.github/workflows/ci.yml and https://github.com/microsoft/Windows-driver-samples/blob/main/.github/workflows/ci-pr.yml .  Those two workflows uses actions that you are updating, so we should also update them for those two files.   Jacob you could take a stab at this and coordinate with @NeoAdonis  .

In hindsight I see similar issue here: https://github.com/microsoft/Windows-driver-samples/actions/runs/8133471231 and here: https://github.com/microsoft/Windows-driver-samples/actions/runs/8133038769 .

NeoAdonis:
I agree with this observation Jakob made. Thanks for the heads-up!

@jacob-ronstadt : Because you have a PR already, could you implement the same change in both ci.yml and ci-pr.yml, please? I could create a separate PR if you prefer that instead.

jacob-ronstadt:
@NeoAdonis @JakobL-MSFT Updated!

JakobL-MSFT (approved):

PR #1136, opened 2024-03-05
NuGet fix for TrEE sample

JakobL-MSFT:
Fix came from Tobi. Thank you!

I tested as follows:
* Build with EWDK NI x64, EWDK GE x64, EWDK DI x64.
* Build with EWDK GE arm64, EWDK DI arm64.
* Build with WDK NuGet GE x64
* Build with WDK NuGet GE arm64
* Build with WDK MSI NI x64

All looked fine.


JakobL-MSFT:
Abandoning.  We will try to do kit change instead.

PR #1153, opened 2024-04-01
Simulated EMI driver sample

tapanansel:
The energy metering functionality in this driver is implemented through the use of the SimEmi virtual child device. Let's break down the key components and their roles:
1. **SimEmiPdoCalculateDataSize**: This function calculates the required size of the child device's PDO (Plug and Play Device Object) data. It takes the number of channels and channel information as input and determines the size of the PDO data structure.
2.	**SimEmiPdoCalculateMetadataSize**: This function calculates the size of the child device's EMI (Energy Metering Interface) metadata structure. It considers the EMI version and the number of channels to determine the size of the metadata structure.
3.	**SimEmiPdoCopyChannelMeasurements**: This function copies the channel measurement data for each channel in the device to the provided output buffer. It updates the channel measurement by calculating the energy gain based on the time elapsed since the last measurement. The measurement data includes the absolute energy and the absolute time of the measurement.
4.	**SimEmiPdoCopyDeviceMetadata**: This function copies the child device's EMI metadata to the provided output buffer. It considers the EMI version and copies the appropriate metadata structure. For EMI version 1, it includes the measurement unit, hardware OEM, hardware model, hardware revision, and metered hardware name. For EMI version 2, it includes the same information along with channel-specific details.
5.	**SimEmiPdoControl**: This function handles device control requests for the EMI child device. It receives IOCTL (Input/Output Control) codes and performs the corresponding actions. For example, it handles requests to retrieve the EMI version, metadata size, metadata, and channel measurements. It calls the respective functions mentioned above to retrieve the required information and copies it to the output buffer.
6.	**SimEmiPdoCreateDevice**: This function creates the PDO for the EMI child device. It initializes the PDO device attributes, assigns hardware and device IDs, sets up device text, creates the PDO device, and initializes the PDO device data. It also creates an IO queue to handle device control requests.
7.	**SimEmiPdoDestroyDevice**: This function cleans up the resources owned by a child device's PDO. It is called when the PDO device is destroyed.
Overall, this driver provides the necessary functions to calculate the size of the PDO data and metadata, copy channel measurements and device metadata, handle device control requests, and create/destroy the PDO device. These functions work together to implement the energy metering functionality for the SimEmi virtual child device.

This driver supports two different versions of the Energy Metering Interface (EMI): EMI version 1 and EMI version 2. Let's discuss the differences between these versions:
1.	**EMI Version 1**: In EMI version 1, each child device has only one channel. The metadata structure for EMI version 1 is represented by the EMI_METADATA_V1 structure. It includes fields such as the measurement unit, hardware OEM, hardware model, hardware revision, and the metered hardware name. The metered hardware name is specific to the channel and is copied from the channel information.
2.	**EMI Version 2**: In EMI version 2, each child device can have multiple channels. The metadata structure for EMI version 2 is represented by the EMI_METADATA_V2 structure. It includes fields for the hardware OEM, hardware model, hardware revision, and the number of channels. Additionally, it includes an array of EMI_CHANNEL_V2 structures, each representing a channel. Each EMI_CHANNEL_V2 structure includes fields for the measurement unit, channel name size, and the channel name itself.
The main difference between EMI version 1 and EMI version 2 is the support for multiple channels in version 2. EMI version 2 allows for more flexibility in representing devices with multiple energy metering channels. This can be useful in scenarios where a single device has multiple energy-consuming components that need to be measured separately.
It's important to note that the driver handles both versions of EMI and provides the necessary functions to calculate the size of the metadata structures, copy channel measurements, and copy device metadata based on the specified EMI version.


simemi/README.MD
```diff
@@ -0,0 +1,16 @@
+The energy metering functionality in this driver is implemented through the use of the SimEmi virtual child device. Let's break down the key components and their roles:
```
NeoAdonis, line 1: File needs a header. The following header might fit your readme:

````
---
page_type: sample
description: "This sample demonstrates the usage of the SimEmi virtual child device for energy metering."
languages:
- cpp
products:
- windows
- windows-wdk
---

# Simulated EMI driver

````

Please change it as you seem fit!

NeoAdonis (changes requested):

PR #1168, opened 2024-05-16
update configuration.dsc.yml for 26100

jacob-ronstadt:


configuration.dsc.yaml
```diff
@@ -4,19 +4,18 @@ properties:
     - resource: Microsoft.WinGet.DSC/WinGetPackage
       id: vsPackage
       directives:
-        description: Install Visual Studio 2022 Community
+        description: Install Visual Studio Community 2022
```
NeoAdonis, line 7: Why was the line changed here?

NeoAdonis (changes requested):
Can you change the target branch to `main`, please? The change can then be merged into `develop` as well.

configuration.dsc.yaml
jacob-ronstadt: That is the output from running the script you provided for generating a new config file.

NeoAdonis: Thanks! I'll check the script later. imho it should be "2022 Community" rather than "Community 2022", but not a big deal by any means.

NeoAdonis:
After changing target branch, it seems that there's an extra commit that shouldn't be merged. Could you rebase the branch to use `main` so that the commit doesn't appear, please?

jacob-ronstadt:
> After changing target branch, it seems that there's an extra commit that shouldn't be merged. Could you rebase the branch to use `main` so that the commit doesn't appear, please?

Which commit are you referring to? [b03f](https://github.com/microsoft/Windows-driver-samples/pull/1168/commits/b03fe77e8157d7a7cc080e689b0485808e733e3b)?

jacob-ronstadt:
Closing to change branch

PR #1172, opened 2024-05-22
update configuration_vsonly.dsc.yaml for 26100

jacob-ronstadt:


JakobL-MSFT:
@jacob-ronstadt .  

This configuration file is - as I best understand - to set up VS to be ready to build drivers with the SDK / WDK NuGet packages.  As such this configuration should not manually install the SDK.  

I should have thought this thru when we spoke and called this out.  I apologize.

jacob-ronstadt:
@JakobL-MSFT Updated!

JakobL-MSFT (approved):

PR #1179, opened 2024-06-07
CI Pipelines build with WDK Nuget Packages

middlemose:
With the WDK now supporting Nuget packages for ease of install it is beneficial to have the CI pipeline also utilize this ability. This will allow easier and faster integration when new WDK versions are published without relying on the runner images to be updated.

- Adds Nuget package capability to the CI pipeline.
- Adds installation of WDK vsix directly into the CI pipeline.
- Updates environment variables for ease of incrementing component versions.

### Pipeline Time Metrics
**BuildAll** (push and pull_request)
Without Vsix Install: ~08m 59s
With Vsix Install: ~38m 59s (optimized ~8m)

**CodeQL** (push and pull_request)
Without Vsix Install: ~1h 1m 43s
With Vsix Install: ~1h 34m 31s (optimized ~1h 16m 2s)

**Average** WDK Vsix Install Time
~21m (optimized ~15s)

**Average** WDK Nuget Restore
~1m 30s

### DIFF
--- a/.github/scripts/Install-Vsix.ps1
+++ b/.github/scripts/Install-Vsix.ps1
@@ -1,0 +1,68 @@
+<#
+
+.SYNOPSIS
+Checks WDK vsix version and downloads and installs as necessary.
+
+#>
+
+[CmdletBinding()]
+param(
+    [bool]$optimize = $false
+)
+
+$root = Get-Location
+
+# launch developer powershell (if necessary)
+if (-not $env:VSCMD_VER) {
+    Import-Module (Resolve-Path "$env:ProgramFiles\Microsoft Visual Studio\2022\*\Common7\Tools\Microsoft.VisualStudio.DevShell.dll")
+    Enter-VsDevShell -VsInstallPath (Resolve-Path "$env:ProgramFiles\Microsoft Visual Studio\2022\*")
+    cd $root
+}
+
+# source environment variables
+. .\Env-Vars.ps1
+
+$version = $env:SAMPLES_VSIX_VERSION
+$uri = $env:SAMPLES_VSIX_URI
+
+function PrintWdkVsix {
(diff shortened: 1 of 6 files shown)


.github/scripts/Install-Vsix.ps1
```diff
@@ -0,0 +1,68 @@
...
+
+[CmdletBinding()]
+param(
+    [bool]$optimize = $false
+)
+
+$root = Get-Location
+
+# launch developer powershell (if necessary)
+if (-not $env:VSCMD_VER) {
```
JakobL-MSFT, line 16: Sorry to be late here, but I would imagine that within GitHub either ($env:VSCMD_VER) is true or it is false.  Always.  Why do you have to code up a conditional manner?  Why not just assert what you believe to be true?

Build-SampleSet.ps1
```diff
@@ -9,6 +9,17 @@ param(
 )
 
 $root = Get-Location
+
+# launch developer powershell (if necessary)
```
JakobL-MSFT, line 13: Similar to a comment for Install-Vsix.ps1.  Sorry to be late here, but I would imagine that anywhere (developer command prompt or GitHub either ($env:VSCMD_VER) is expected to be true or it is expected to be false. Why do you have to code up a conditional manner? Why not just assert what you believe to be true?

JakobL-MSFT (approved):
Approved with non-blocking comments.

Build-SampleSet.ps1
middlemose: The thought behind this logic is when a user executes the script(s) locally and the environment may or may not be a developer shell already.

.github/scripts/Install-Vsix.ps1
middlemose: Thanks for comment here. Will consider changing in the future.

PR #1196, opened 2024-07-08
update with codel fix

taj0207:
fix codel's error

### DIFF
--- a/network/wwan/cxwmbclass/utils.cpp
+++ b/network/wwan/cxwmbclass/utils.cpp
@@ -248,7 +248,7 @@ MbbBusGetStat(__in MBB_BUS_HANDLE BusHandle, __in USHORT StatIndex, __out ULONGL
 
     if (NT_SUCCESS(Status))
     {
-        if (BytesTransfered < sizeof(Value))
+        if (BytesTransfered < sizeof(*Value))
         {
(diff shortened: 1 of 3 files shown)


network/wwan/cxwmbclass/utils.cpp
```diff
@@ -248,7 +248,7 @@ MbbBusGetStat(__in MBB_BUS_HANDLE BusHandle, __in USHORT StatIndex, __out ULONGL
 
     if (NT_SUCCESS(Status))
     {
-        if (BytesTransfered < sizeof(Value))
+        if (BytesTransfered < sizeof(*Value))
```
ajkann, line 251: Value seem to be an out value 
__out ULONGLONG* Value
Is it meant to be dereferenced ?

taj0207: At this point, there should be value pointed after SendSyncControlCommand success. Here is to check the BytesTransfered larger or equal to the dereferenced value's size.

taj0207: Yes, it is meant to be dereferenced, the bufferlength put into the controlcommand req is also *value. Only sizeof (ULONGLONG ) as input buffer lenght.

PR #1199, opened 2024-07-15
Update userscan.c

wqreytuk:
shouldn't this being abortion port？


cgallred (changes requested):
Should be "abort" port, not "abortion" port.

PR #1212, opened 2024-09-05
Update Capture.cpp

gerritsMSFT:
Add Return check inside NT_SUCCESS macro


LPBourret (approved):

middlemose:
@gerritsMSFT If you update your branch to latest `develop` and then rerun checks they should then pass.

PR #1218, opened 2024-09-19
No longer install VSIX as part of CI pipeline

JakobL-MSFT:


JakobL-MSFT:
Abandoning this PR. Pending an alternative version from @middlemose

PR #1222, opened 2024-09-26
[usb/usbview] Fix memory leak when handling enumeration error cases

dkala97:
When handling error cases in the `EnumerateHub` function, the following objects aren't freed:
- ConnectionInfoV2
- PortConnectorProps
- DevProps
- hubCapabilityEx

When handling error cases in the `EnumerateHubPorts` function, when a port is not a hub, the following objects aren't freed:
- stringDescs
- DevProps

### DIFF
--- a/usb/usbview/enum.c
+++ b/usb/usbview/enum.c
@@ -951,6 +951,16 @@ EnumerateHubError:
         FREE(ConnectionInfo);
     }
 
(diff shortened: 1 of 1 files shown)


dkala97:

dkala97:

dkala97:
Kindly reminder

PR #1228, opened 2024-10-31
Add a KASAN demo sample

JohnMcSandwich:
Add a sample driver that demonstrates how KASAN catches certain bugs.


JakobL-MSFT (approved):
We need to afterwards also update https://github.com/microsoft/Windows-driver-samples/blob/main/.github/CODEOWNERS but let's not wait for that.

PR #1252, opened 2025-01-16
fix WFPSampler log info

kouzhudong:


kouzhudong:

PR #1294, opened 2025-06-23
Fixes and updates for StorageCore driver sample components

irvinemaddox:
Fix 1: cpp/drivers/opaque-mdl-use codeql query. Fix for opaque MDL field accesses in classpnp and disk. Use recommended API.
Fix 2: cpp/uninitialized-local codeql query. Fix for uninitialized variables during access.
Fxi 3: cpp/conditionally-uninitialized-variable codeql rule. Initialize variables that are passed by reference to a function if the status of the function is not checked.
Fix 4: Update INF file for lsi_u3 driver to use $ARCH$ for platform-specific stamping via Stampinf.


1Sisyphus:
@microsoft-github-policy-service agree company="Microsoft"

PR #1315, opened 2025-07-30
Fix the wdk version in EWDK

5an7y-Microsoft:
Build-sample set was printing an error message when trying to find the WDK version while using the EWDK.

# Behavior without changes

image

# Behavior with changes

image


5an7y-Microsoft:
@microsoft-github-policy-service agree company="Microsoft"

JakobL-MSFT (approved):

PR #1317, opened 2025-07-31
Updated Sample Driver Enum

Gshri1021:
Currently the Serial sample driver sets wakeSettings.UserControlOfWakeSettings to an incompatible enum type. The value was updated to be WakeDoNotAllowUserControl so that the driver can build correctly instead.

Screenshot 2025-07-31 131820


AndrewYangMSFT (approved):

AndrewYangMSFT:
CodeQL failed due to its config issue. Ignoring the failure as the change has been verified locally.

Issue #60, opened 2016-05-11
license for contributions


jlovejoy:
Can you please clarify or confirm whether contributions made to this project are taken in under the same/outbound license, that is: Microsoft Public License (MS-PL)?

As a side note, it looks like all your source files have "all rights reserved" with no indication of the license that applies, you might want to add a notice in the source files MS-PL applies (best practice, a la ASF and FSF projects).  :)

Thanks!

wm1:
There is a LICENSE file.

PR #1335, opened 2025-11-10
Add SPDX license identifiers to source files

rwmjones:
SPDX (https://spdx.org/licenses/) is a commonly used system for annotating source files with the intended license.  In WDK, the top level LICENSE file describes the license as MS-PL, but individual files are less clear about the exact license.  This has caused confusion in the past
(https://github.com/microsoft/Windows-driver-samples/issues/60).  By annotating files with SPDX the license can be clarified by the addition of a single line.

This change was done largely using the script below, followed by some light hand editing:
```
perl -i -e 'while (<>) { if ($. <= 5 && m/(.*)Copyright.*Microsoft/) { print; print "$1SPDX-License-Identifier: MS-PL\r\n"; } else { print } } continue { close ARGV if eof }' -- $( git ls-files )
```


rwmjones:
@microsoft-github-policy-service agree company="Red Hat, Inc."

JakobL-MSFT:
I appreciate the effort, but, frankly, as I balance it out, I am not sure this is a good change or an important area to improve. 

We are with this change emphasizing explicitness at the cost of redundancy and compactness.   Frankly, I am more inclined to go to in different direction and delete lines referencing root of repo.

I may be wrong in this balance, and I will keep an open mind to be proven wrong.

rwmjones:
Link to discussion on Fedora legal which led to this PR: https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/ZSVCFJFVMLTODFKGV352TYHH5ZQODJXD/

JakobL-MSFT:
Thank you for sharing discussion.  I am intrigued.  I will ask around internally at MSFT to see what the consensus is.

PR #1385, opened 2026-05-19
Fix country names to readable Chinese in rtchnlplan.h

andremueiot:
1. Country name strings were updated from garbled mojibake text to readable Chinese text in the main country/domain table around rtchnlplan.h:102.
2. Matching comment strings in the country enum section were updated the same way around rtchnlplan.h:261.
3. No functional code logic changed:
    - Enum identifiers like RT_CTRY_AL, RT_CTRY_US, etc. are unchanged.
    - Domain code fields (2G_WORLD, 5G_ETSI1, etc.) are unchanged.
    - This is primarily a text/content cleanup plus minor spacing/alignment edits.
 
With these changes the writing in Chinese shows up properly in Visual Studio.


andremueiot:
@keith-horton , FYI.

keith-horton:
> @keith-horton , FYI.

The web view of this change makes it look like they are practically identical :(
Are they different if I use another tool to compare them?

andremueiot:
> > @keith-horton , FYI.
> 
> The web view of this change makes it look like they are practically identical :( Are they different if I use another tool to compare them?

I was also surprised by seeing these changes shown as similar. But note that the coloring still shows up, so clearly the tool understands that there are differences.

When I have the file open in VS Code and switch from main to this branch I can clearly see the changes. I'm not sure why the web interface is not showing the contents that we saw in that other PR which was removing the trailing spaces. Maybe it's taking hints from the new changes?

Anyways, I'm convinced that these changes are good :-)

andremueiot:
The root cause for all the confusion here is that the file does not have a BOM. On Windows, different tools end up viewing the data differently. The best fix would be to add a BOM to this file. I'll leave that to the owners of the file.

PR #1392, opened 2026-06-09
Fix to use CPP20 for compiler

gerritsMSFT:
STD 17 used experimental coroutine. STD 20 uses standard C++. WinRT will inherit the coroutine automatically, so updating to STD 20 will fix this issue.


deniss-msft (approved):

general/SimpleMediaSource/MediaSource/MediaSource.vcxproj
```diff
@@ -99,7 +98,7 @@
       <EnablePREfast>false</EnablePREfast>
       <ExceptionHandling>Sync</ExceptionHandling>
       <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
```
NaveenThum, line 101: Is c++latest allowed here?

deniss-msft: I don't hear good things about using latest. It feels like it will increase upkeep, not reduce it: https://developercommunity.visualstudio.com/t/11089187 or 
image(from https://github.com/microsoft/STL/issues/4876)

deniss-msft (approved):

5an7y-Microsoft (approved):

PR #1399, opened 2026-06-30
wificx: enable and test the two-instance ENL data path

krish2718:
## Summary

Enables and documents the **two-instance data path** for the WiFiCx sample:
two emulated stations now scan, connect, and exchange real traffic (ping)
over the Emulated Network Link (ENL) — the Windows analogue of Linux
`mac80211_hwsim` — with no physical Wi-Fi hardware.

Bringing this up end-to-end required fixing several gaps in the control and
data paths, a NetAdapterCx config bug, and the build/INF packaging, plus a
deterministic deploy/test workflow.

## What's included

- **`netvadapter`: build fix** — build with `/t:Rebuild` so an incremental
  build can't reuse a stale catalog (drvcat signature mismatch at install).
- **`netvadapter`: config default handling** — `NetvAdapterReadConfiguration`
  stored values before validating and never wrote the fallback back, so a
  missing/out-of-range `MACLastByte` (table default `0`, rejected as `< 1`)
  could fail adapter start. Validate first, store once, default/min = 1.
- **`wificx`: connect → data-path activation** — start the adapter
  disconnected (so the framework runs CONNECT, not ROAM), always report
  `Software=On` for the radio-less loopback adapter (WS2025 otherwise never
  issues `WDI_TASK_CONNECT`), and complete unhandled `WDI_TASK_*` commands
  with M3+M4 instead of M3 only.
- **`wificx`: AP-relay receive fix** — both stations associate to the same
  fake AP BSSID; with no AP to forward between them, ENL copied each
  station's "to-DS" frame verbatim, so `nwifi` on the peer dropped it (RA was
  the BSSID, wrong DS direction). ENL now rewrites the forwarded 802.11
  header into the "from-DS" frame the peer accepts, exactly as an AP would.
- **`wificx`: deploy/test scripts** — `deploy.ps1` builds, signs, installs
  two instances and pairs them deterministically by writing `MACLastByte` to
  each adapter's **software (Class) key** (the location NetAdapterCx reads)
  after a clean-slate devnode removal — no manual step or reboot. `run.ps1`
  connects both, assigns IPs, seeds static neighbors (same-host
  duplicate-address ARP workaround), and pings both ways.
- **Docs** — `README_EWDK.md` (EWDK build/deploy/test) and a `README.md`
  note on the ENL relay role.
- **Debug tracing** — all `DbgPrintEx` diagnostics are isolated in the final
  commit behind off-by-default switches (`NETV_DATAPATH_DEBUG` /
  `WIFICX_DATAPATH_DEBUG`); they compile out of a normal build.

## Testing

On a Windows Server 2025 VM (EWDK build):

```powershell
.\network\wlan\wificx\deploy.ps1   # builds, installs, pairs (MACLastByte 1 & 2)
.\network\wlan\wificx\run.ps1      # connects both, pings both directions
```

Both adapters scan, connect to the open SSID, and report
`PASS: Bidirectional ping succeeded` between `192.168.100.1` and
`192.168.100.2`.

## Notes

- Pure software simulation; no Wi-Fi hardware required.
- Both stations share one host/subnet, so `run.ps1` seeds static neighbor
  entries to bypass duplicate-address ARP suppression.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/network/netadaptercx/netvadapter/build.cmd
+++ b/network/netadaptercx/netvadapter/build.cmd
@@ -21,5 +21,5 @@ set PLAT=%~2
 if "%CONFIG%"=="" set CONFIG=Debug
 if "%PLAT%"=="" set PLAT=x64
 
-msbuild "%~dp0netvadapter.sln" /t:Build /p:Configuration=%CONFIG% /p:Platform=%PLAT% /p:UMDF_VERSION_MINOR=33 /p:UMDF_MINIMUM_VERSION_REQUIRED=33 /m
+msbuild "%~dp0netvadapter.sln" /t:Rebuild /p:Configuration=%CONFIG% /p:Platform=%PLAT% /p:UMDF_VERSION_MINOR=33 /p:UMDF_MINIMUM_VERSION_REQUIRED=33 /m
 endlocal
--- a/network/netadaptercx/netvadapterlibrary/code/adapter.cpp
+++ b/network/netadaptercx/netvadapterlibrary/code/adapter.cpp
@@ -307,7 +307,11 @@ EvtSetReceiveFilter(
     )
 {
     NetvAdapter* adapter = NetvAdapterGetContextFromWDFObject(NetAdapter);
-
+#if defined(_KERNEL_MODE) && defined(NETV_DATAPATH_DEBUG)
+    DbgPrintEx(DPFLTR_IHVDRIVER_ID, DPFLTR_ERROR_LEVEL,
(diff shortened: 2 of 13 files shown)


krish2718:
@microsoft-github-policy-service agree [company="{Dotstar systems}"]

krish2718:

krish2718:
@youyanggit please approve the workflows and also review.

